PR #6, opened 2017-01-12
added AbstractEnvironmentLoaderListener to enable cdi injectable realms

voons:


pom.xml
```diff
@@ -107,10 +118,19 @@
...
+            <artifactId>javax.servlet-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
             <groupId>javax.validation</groupId>
             <artifactId>validation-api</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.apache.shiro</groupId>
+            <artifactId>shiro-web</artifactId>
```
bannmann: like dynamodb, the shiro dependency should be optional.

PR #17, opened 2017-03-10
added "aws.region" system property for configuration of the aws region

voons:


src/main/java/com/github/mizool/aws/dynamodb/DynamoDbMapperProducer.java
```diff
@@ -30,7 +29,7 @@
     public DynamoDbMapperProducer()
     {
         client = new AmazonDynamoDBClient();
-        client.withRegion(Regions.EU_CENTRAL_1);
+        client.withRegion(new Configuration().getAwsRegion());
```
bannmann: any reason not to go with `@Inject` here?

PR #21, opened 2017-03-20
added mail classes

voons:


src/main/java/com/github/mizool/Clock.java
```diff
@@ -0,0 +1,11 @@
+package com.github.mizool;
```
bannmann: license boilerplate is missing

PR #23, opened 2017-03-21
stacktrace details must be whitelisted in order to be displayed

voons:


src/main/java/com/github/mizool/rest/errorhandling/DefaultWhiteList.java
```diff
@@ -32,6 +33,7 @@
 import com.github.mizool.exception.UnprocessableEntityException;
 import com.github.mizool.exception.UnsupportedHttpFeatureException;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 
 @MetaInfServices
 public class DefaultWhiteList implements WhiteList
```
bannmann: The naming and structure is weird, let's use the opportunity to clean this up:

- Create a class `WhiteListEntry`
    - Add the fields `String exceptionClassName`, `int statusCode`, both final and with `@Getter`
    - Add the field `boolean shouldIncludeDetails`, also final, but without `@Getter`
        - Implement a getter called `getShouldIncludeDetails()` (as lombok would generate an `is...()` getter which doesn't fit)
    - Have lombok generate an `AllArgsConstructor`
    - In order to make `shouldIncludeDetails` an optional constructor parameter, write an additional constructor that sets `shouldIncludeDetails` to `false`.
- Refactor `getEntries()` to return `Set<WhiteListEntry>`
- Refactor `ExceptionCatalog`:
    - iterate the set returned by `getEntries()` and populate an internal `Map<String, WhitelistEntry>`
    - remove `showStacktrace()`
    - change `lookup()` to return `Optional<WhiteListEntry>`
- Change `ErrorHandler` and `DefaultWhiteList` accordingly

PR #24, opened 2017-03-21
AuthenticationRejectedException now inherits from Shiro Authenticatio…

voons:
Shiro was throwing a warning that exceptions during the login process should inherit from the AuthenticationException


voons:
found a better way

PR #27, opened 2017-03-23
added dynamoDB low level api producer

voons:


src/main/java/com/github/mizool/aws/dynamodb/AmazonDynamoDbProducer.java
```diff
@@ -0,0 +1,27 @@
+package com.github.mizool.aws.dynamodb;
```
bannmann: license boilerplate

PR #30, opened 2017-04-27
added Date converter for ZonedDateTime

voons:


voons:
will be done as part of another branch

PR #34, opened 2017-05-12
added locked exception

voons:


core/src/main/java/com/github/mizool/core/exception/LockedException.java
```diff
@@ -0,0 +1,43 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package com.github.mizool.core.exception;
+
+/**
+ * Thrown when trying to manipulate a locked entity
+ */
+public class LockedException extends RuntimeException
```
bannmann: Rename to `LockedEntityException`

PR #35, opened 2017-06-06
added CacheHeaderFilter

voons:


bannmann:
the build failed, probably due to the move from 0.2-SNAPSHOT to 0.3-SNAPSHOT?

technology/web/src/test/java/com/github/mizool/technology/web/TestCacheHeaderFilter.java
```diff
@@ -0,0 +1,115 @@
...
+    {
+        verify(servletResponse, never()).setHeader(eq("Cache-Control"), anyString());
+        verify(servletResponse, never()).setHeader(eq("Pragma"), anyString());
+        ArgumentCaptor<Long> argumentCaptor = ArgumentCaptor.forClass(Long.class);
+        verify(servletResponse).setDateHeader(eq("Expires"), argumentCaptor.capture());
+        assertThat(argumentCaptor.getValue()).isGreaterThan(0);
+    }
+
+    @Test
+    public void testWithIndexHtml() throws IOException, ServletException
```
bannmann: rework the tests to use only two test methods with data providers. inline the constants.
keep the `verifyNotCached` and `verifyCachedForever` methods separate, but move them below the test methods.

technology/web/src/main/java/com/github/mizool/technology/web/CacheHeaderFilter.java
```diff
@@ -0,0 +1,94 @@
...
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+public class CacheHeaderFilter implements Filter
+{
+    private static final Pattern STATIC_ELEMENTS_PATH_PATTERN = Pattern.compile(
+        "^/(webjars|\\d+(\\.\\d+)*(-[\\dA-Za-z]+)?)/.+$");
```
bannmann: hmm, the `(-[\\dA-Za-z]+)?)` looks like it was meant to catch `-SNAPSHOT` as well, but fails to do that for some reason I can't easily make out.

as it makes more sense to disable caching of snapshot artifacts, we should simplify the pattern to `^\\/(webjars|\\d+(\\.\\d+)*)\\/`

also, add a link http://regexr.com/3g43r

note that this will remove support for enabling caching for versions like `1.0-beta3`, which is okay as we don't use those.

bannmann: okay, as the proposed version breaks the tests (even if replacing `\\/` with the original `/`) and because I now found the reason it doesn't catch snapshots (separate `path.contains` call in line 64), we should just keep this as is.

PR #37, opened 2017-06-14
upgraded jackson due to security problems with old version

voons:


pom.xml
```diff
@@ -101,24 +102,24 @@
                 <!-- force dependency convergence. 2.8.4 is intended for jersey-media-json-jackson 2.25 -->
```
bannmann: change all these comments to `required for dependency convergence with transitive dependencies of jersey-media-jackson`

pom.xml
```diff
@@ -66,6 +66,7 @@
         <aws.sdk.version>1.11.75</aws.sdk.version>
         <cassandra.driver.version>3.2.0</cassandra.driver.version>
         <encoding>UTF-8</encoding>
+        <jackson.version>2.8.9</jackson.version>
```
bannmann: * add comment: `to fix a vulnerability, we use jackson 2.8.9 instead of 2.8.4 which would match jersey 2.25`
* move `jersey-media-json-jackson` version to a new property `jersey.version`

PR #39, opened 2017-07-27
added validator for iso country codes

voons:


core/src/main/java/com/github/mizool/core/validation/CountryCodeValue.java
```diff
@@ -0,0 +1,44 @@
...
+import java.lang.annotation.Target;
+
+import javax.validation.Constraint;
+import javax.validation.Payload;
+
+@Target({ METHOD, FIELD, ANNOTATION_TYPE, CONSTRUCTOR, PARAMETER })
+@Retention(RUNTIME)
+@Constraint(validatedBy = CheckCountryCodeValue.class)
+@Documented
+public @interface CountryCodeValue
```
bannmann: the "value" suffix is only used when there is a different type, as with `EnumValue`. we don't use it when it's about a regular string, such as `Email` (not `EmailValue`).

for consistency, rename this and the check to `CountryCode`.

core/src/main/java/com/github/mizool/core/validation/CountryCode.java
```diff
@@ -32,9 +32,9 @@
 
 @Target({ METHOD, FIELD, ANNOTATION_TYPE, CONSTRUCTOR, PARAMETER })
 @Retention(RUNTIME)
-@Constraint(validatedBy = CheckCountryCodeValue.class)
+@Constraint(validatedBy = CheckCountryCode.class)
 @Documented
-public @interface CountryCodeValue
+public @interface CountryCode
 {
     String message() default "{com.github.mizool.core.validation.CountryCodeValue.message}";
```
bannmann: adapt the string

PR #41, opened 2017-08-08
added ServiceUnavailableException

voons:


core/src/main/java/com/github/mizool/core/exception/ServiceUnavailableException.java
```diff
@@ -0,0 +1,45 @@
...
+    {
+        super(message, cause);
+    }
+
+    public ServiceUnavailableException(Throwable cause)
+    {
+        super(cause);
+    }
+
+    public ServiceUnavailableException(
```
bannmann: we don't (shouldn't) include this constructor in our exceptions

PR #46, opened 2017-09-19
added RuleViolationException

voons:


core/src/main/java/com/github/mizool/core/exception/RuleViolationException.java
```diff
@@ -0,0 +1,49 @@
...
+    {
+        super(message);
+
+        if (ruleViolations == null)
+        {
+            this.ruleViolations = null;
+        }
+        else
+        {
+            this.ruleViolations = new HashSet<>(ruleViolations);
```
bannmann: this would allow manipulating the set as `getRuleViolations()` returns it without protecting it. make an `ImmutableSet` copy instead of using `HashSet`.

core/src/main/java/com/github/mizool/core/exception/RuleViolationException.java
```diff
@@ -0,0 +1,49 @@
...
+{
+    private final Set<RuleViolation> ruleViolations;
+
+    public RuleViolationException(String message, Set<RuleViolation> ruleViolations)
+    {
+        super(message);
+
+        if (ruleViolations == null)
+        {
+            this.ruleViolations = null;
```
bannmann: this is not useful, we can at least expect an empty set. annotate `ruleViolations` parameter with lombok's `@NonNull`.

core/src/main/java/com/github/mizool/core/rest/errorhandling/ErrorHandler.java
```diff
@@ -75,6 +77,10 @@ else if (isAssignable(ClientErrorException.class, cursor))
             {
                 result = handleClientError((ClientErrorException) cursor);
             }
+            else if (isAssignable(RuleViolationException.class, cursor))
```
bannmann: move this up so it's directly below the `ConstraintViolationException` stuff to which it is conceptually similar. incidentally, the order of `if` branches then conforms to the order of the methods they call.

core/src/main/java/com/github/mizool/core/rest/errorhandling/ErrorHandler.java
```diff
@@ -151,6 +169,12 @@ private void recordViolation(ConstraintViolation<?> violation, ListMultimap<Stri
         target.put(propertyName, errorDto);
     }
 
+    private void recordRuleViolation(RuleViolation violation, ListMultimap<String, ErrorDto> target)
```
bannmann: move this up so it follows `handleRuleViolationError()`.

bannmann: for consistency, also do the following:

- adapt the position of `recordViolation()` accordingly
- rename `recordViolation()` to `recordConstraintViolation()`
- rename `handleValidationError()` to `handleConstraintViolationError()`

core/src/main/java/com/github/mizool/core/exception/RuleViolationException.java
```diff
@@ -0,0 +1,49 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package com.github.mizool.core.exception;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class RuleViolationException extends RuntimeException
```
bannmann: `RuleViolationException` still needs to be added to the whitelist!

bannmann: never mind, it doesn't.

PR #56, opened 2018-01-05
added jdk 8 time codecs to cassandra cluster

voons:


core/src/main/java/com/github/mizool/core/converter/ZonedDateTimeConverter.java
```diff
@@ -1,21 +1,23 @@
 /**
- *  Copyright 2017 incub8 Software Labs GmbH
- *  Copyright 2017 protel Hotelsoftware GmbH
+ * Copyright 2017 incub8 Software Labs GmbH
```
bannmann: please change the year to `2017-2018` in every existing file you touch, for both copyright lines.

PR #65, opened 2018-02-16
Added BlockingAsynchronousFunctionDecorator as a replacement for the …

schroenser:
…BlockingAsynchronoausTaskController.

### DIFF
--- a/core/src/main/java/com/github/mizool/core/function/BlockingAsynchronousFunctionDecorator.java
+++ b/core/src/main/java/com/github/mizool/core/function/BlockingAsynchronousFunctionDecorator.java
@@ -1,0 +1,94 @@
+/**
+ * Copyright 2018 incub8 Software Labs GmbH
+ * Copyright 2018 protel Hotelsoftware GmbH
+ *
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/github/mizool/core/function/BlockingAsynchronousFunctionDecorator.java
```diff
@@ -0,0 +1,94 @@
...
+
+    private final Function<T, ListenableFuture<R>> target;
+    private final int taskLimit;
+
+    private final AtomicInteger runningTasks = new AtomicInteger();
+    private final Object semaphore = new Object();
+    private final CompletionListener completionListener = new CompletionListener();
+
+    /**
+     * Applies the {@code target} {@link java.util.function.Function} to the given argument. Application is delayed if
```
bannmann: Avoid the term "delayed/resumed", be specific. Replace the second sentence with 
```
If the {@code taskLimit} is reached, the calling thread is blocked until one or more tasks complete.
```

and adapt the class comment accordingly.

PR #68, opened 2018-02-23
Feature/mandatory validation annotations

voons:


core/src/main/java/com/github/mizool/core/validation/CheckCountryCode.java
```diff
@@ -1,41 +1,53 @@
...
+        mandatory = countryCode.mandatory();
     }
 
     @Override
-    protected boolean isNonNullNonEmptyStringValid(
-        String validationObject, ConstraintValidatorContext constraintValidatorContext)
+    public boolean isValid(String validationObject, ConstraintValidatorContext constraintValidatorContext)
     {
-        return COUNTRY_CODES.contains(validationObject);
+        boolean valid = false;
```
bannmann: refactor this so `isValid()` only contains `return isNullButOptional(validationObject) || isValidValue(validationObject);`. do the same for the other checks.

this way, the reader does not have to parse a mixture of negations and if/else structure.

core/src/main/java/com/github/mizool/core/validation/CheckCountryCode.java
```diff
@@ -1,41 +1,53 @@
 /**
- *  Copyright 2017 incub8 Software Labs GmbH
- *  Copyright 2017 protel Hotelsoftware GmbH
+ * Copyright 2017 incub8 Software Labs GmbH
```
bannmann: update license years in all files touched by this PR

core/src/main/java/com/github/mizool/core/validation/Nullable.java
```diff
@@ -0,0 +1,38 @@
...
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/**
+ * This annotation is the opposite of {@link javax.validation.constraints.NotNull} and is intended to document that a field is optional
+ */
+@Target({ METHOD, FIELD, ANNOTATION_TYPE, CONSTRUCTOR, PARAMETER })
```
bannmann: only keep `FIELD`, the others don't make sense for our use case

voons: we should keep `PARAMETER` as well in order to annotate rest method parameters.

core/src/main/java/com/github/mizool/core/validation/CheckCountryCode.java
```diff
@@ -1,41 +1,54 @@
...
 
     @Override
-    protected boolean isNonNullNonEmptyStringValid(
-        String validationObject, ConstraintValidatorContext constraintValidatorContext)
+    public boolean isValid(String validationObject, ConstraintValidatorContext constraintValidatorContext)
+    {
+        return isNullButOptional(validationObject) || isValidValue(validationObject);
+    }
+
+    private Boolean isNullButOptional(String validationObject)
```
bannmann: using a `Boolean` wrapper is imprecise and unnecessary as the condition will always be either `true` or `false`. always use `boolean` unless there's a reason to use the wrapper (e.g. fields in DTOs that should not be `false` when left out).

do this for the other classes as well.

PR #69, opened 2018-02-23
Feature/configurable cassandra connection

voons:


technology/cassandra/src/main/java/com/github/mizool/technology/cassandra/ClusterProducer.java
```diff
@@ -33,6 +36,12 @@
 class ClusterProducer
 {
     private static final String CASSANDRA_CONTACT_POINTS_PROPERTY_NAME = "cassandra.contactpoints";
+    private static final String
+        CASSANDRA_MAX_REQUESTS_PER_CONNECTION_LOCAL_PROPERTY_NAME
+        = "cassandra.max.requests.per.connection.local";
```
bannmann: when using the period, it must separate logical hierarchy levels. change this to `cassandra.maxRequestsPerConnection.local` and adapt the other name accordingly.

Note: yes, using camel case here is inconsistent with `cassandra.contactpoints`, but that's not worth breaking the API, and restricting to lower case feels too ugly.

technology/cassandra/src/main/java/com/github/mizool/technology/cassandra/ClusterProducer.java
```diff
@@ -77,6 +87,26 @@ private void registerJdk8TimeCodecs(Cluster cluster)
             .register(LocalTimeCodec.instance);
     }
 
+    private PoolingOptions getPoolingOptions()
+    {
+        PoolingOptions poolingOptions = new PoolingOptions();
+
+        int maxRequestsPerConnectionLocal = Integer.parseInt(System.getProperty(
```
bannmann: loading the default value makes this code unnecessarily complex. furthermore, making an effort to re-set the default feels nonsensical and like cargo culting.

we should either hard-code the default value in our code or skip the setter call if the respective system property wasn't set. I prefer the latter.

PR #71, opened 2018-03-02
Fix misleading constant name "GENERIC_FIELD_KEY" in ErrorDto

bannmann:


core/src/main/java/com/github/mizool/core/rest/errorhandling/ErrorDto.java
```diff
@@ -26,11 +26,16 @@
 @RequiredArgsConstructor(access = AccessLevel.PACKAGE)
 public class ErrorDto
 {
-    public static final String GENERIC_FIELD_KEY = "generic";
+    private static final String GENERIC_ERROR_ID = "generic";
```
schroenser: Leave this `public` as code outside of mizool accesses the value.

schroenser: While you're at it, also make `ErrorHandler.GLOBAL_PROPERTY_KEY` public as we already have a part in the code that duplicates that value instead of just referencing the original constant.

PR #72, opened 2018-03-12
Extract record methods

lesmou:


pom.xml
```diff
@@ -44,6 +44,18 @@
             <organization>incub8 Software Labs</organization>
             <organizationUrl>http://www.incub8.de</organizationUrl>
         </developer>
+        <developer>
+            <name>Sven Haberer</name>
```
bannmann: I know it's not obvious from the existing entries, but I'd like to sort these "mechanically" based on the entire `<name>`, not "smart" on the last name, which may not even work for some people (with no or multiple last names).

core/src/main/java/com/github/mizool/core/rest/errorhandling/ErrorHandler.java
```diff
@@ -42,18 +40,24 @@
 
     private final ExceptionCatalog exceptionCatalog;
     private final ErrorMapper errorMapper;
```
bannmann: Okay, that's clearly better than keeping everything in one class.

However, reading the class now, I feel having a method like `handleWhitelistedException()` in this class, but having that method call `createExceptionParameters()` and `createErrorMessageDto()` in another class feels like a somewhat arbitrary separation of classes.

Let's continue moving out stuff of this class:
* `handleWhitelistedException()`, `handleClientError()` (incl `private determineErrorClass()`) and `handleUndefinedError()` move to `ErrorMapper`
* `handleConstraintViolationError()` moves to `ConstraintViolationMapper`
* `handleRuleViolationError()` moves to `RuleViolationMapper`

PR #74, opened 2018-03-16
Feature/blockingasynchronousfunctiondecorator does not do the right thing

schroenser:
As @bannmann was heavily involved in implementing this, we decided to have @JanDoerrenhaus have a look at it instead.
Note that this branch is based on the branch from pull request #73. Land that one first to make the review easier.


core/src/main/java/com/github/mizool/core/concurrent/ListenableFutureCollector.java
```diff
@@ -0,0 +1,198 @@
...
+                }
+                semaphore.notifyAll();
+            }
+        }
+    }
+
+    private final int maximumConcurrentFutures;
+    private final Object semaphore = new Object();
+
+    private int runningFutures;
```
JanDoerrenhaus: Use an AtomicInteger instead. Due to the semaphore, the regular int should not pose any problems with its ++ and -- operators, but AtomicIntegers incrementAndGet/decrementAndGet methods will make the intent more clear, I think.

core/src/main/java/com/github/mizool/core/concurrent/ListenableFutureCollector.java
```diff
@@ -0,0 +1,199 @@
...
+                setResultValueIfAllFuturesFinished();
+                return result;
+            }
+        };
+    }
+
+    @Override
+    public Set<Characteristics> characteristics()
+    {
+        return EnumSet.of(Characteristics.CONCURRENT, Characteristics.UNORDERED);
```
bannmann: I now think we should not claim to be `CONCURRENT` as it gives a false impression.

[As explained here](https://stackoverflow.com/a/29920375/7641), non-`CONCURRENT` collectors can still be used from multiple threads, which would mean we can no longer guarantee that the limit is kept. But I think that's okay, provided we add a note about this in the docs.

PR #77, opened 2018-03-22
Added collector for guava's immutable collections.

schroenser:


core/src/main/java/com/github/mizool/core/GuavaCollectors.java
```diff
@@ -0,0 +1,156 @@
...
+        Function<? super T, ? extends K> keyMapper, Function<? super T, ? extends U> valueMapper)
+    {
+        return new CollectorImpl<>(
+            ImmutableMap::builder,
+            (builder, element) -> builder.put(keyMapper.apply(element), valueMapper.apply(element)),
+            (thisBuilder, otherBuilder) -> thisBuilder.putAll(otherBuilder.build()),
+            ImmutableMap.Builder::build);
+    }
+
+    public static <T, K, U> Collector<T, ImmutableMultimap.Builder<K, U>, ImmutableMultimap<K, U>> toImmutableMultimap(
```
bannmann: Do we ever use `ImmutableMultimap` directly instead of either `ImmutableListMultimap` or `ImmutableSetMultimap`? As quoted below, [Guava does not recommend this](https://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/ImmutableMultimap.html), so we should probably not offer a collector for that type.

> *Warning:* avoid _direct_ usage of `ImmutableMultimap` as a type (as with `Multimap` itself). Prefer subtypes such as `ImmutableSetMultimap` or `ImmutableListMultimap`, which have well-defined `equals(java.lang.Object)` semantics, thus avoiding a common source of bugs and confusion.

PR #80, opened 2018-05-03
Threads utility class.

JanDoerrenhaus:


core/src/main/java/com/github/mizool/core/Threads.java
```diff
@@ -0,0 +1,22 @@
+package com.github.mizool.core;
```
bannmann: move to `com.github.mizool.core.concurrent`

schroenser:
Please also use this class in `ListenableFutureCollector` and `BlockingAsynchronousTaskController`.

core/src/main/java/com/github/mizool/core/Threads.java
bannmann: oh, and add the license boilerplate

bannmann:
@schroenser, for what? These classes don't invoke `sleep()`.

core/src/main/java/com/github/mizool/core/concurrent/Threads.java
```diff
@@ -0,0 +1,51 @@
...
+{
+    public void sleep(int milliSeconds)
+    {
+        try
+        {
+            Thread.sleep(milliSeconds);
+        }
+        catch (InterruptedException e)
+        {
+            Thread.currentThread().interrupt();
```
bannmann: move these two lines to a shared private method `rethrowInterrupt()`

core/src/main/java/com/github/mizool/core/concurrent/ListenableFutureCollector.java
```diff
@@ -150,15 +150,7 @@ private void waitWhileCapacityReached()
...
-            try
-            {
-                semaphore.wait();
-            }
-            catch (InterruptedException e)
-            {
-                Thread.currentThread().interrupt();
-                throw new UncheckedInterruptedException(e);
-            }
+            Threads.wait(semaphore);
```
bannmann: also use it on `BlockingAsynchronousTaskController` - no reason to leave that code ugly just because it's deprecated

schroenser:
@bannmann I thought that the wait call on the semaphores was similar in nature... but yeah... this is about `Threads.sleep()`. My bad.

PR #82, opened 2018-05-17
Added mockIterable to MockitoExtensions

Yombu:


core/src/test/java/com/github/mizool/core/MockitoExtensions.java
```diff
@@ -18,6 +18,9 @@
```
bannmann: update copyright year to `2017-2018`

bannmann: right, I forgot about the `NOTICE` file. however, my comment referred to the `MockitoExtensions` boilerplate comment.

Yombu: My bad. Didn't see this because IntelliJ wraps the boilerplate code neatly to a single (unobtrusive) line.

PR #89, opened 2018-06-25
added system property that can be used to set the client side time out

codingWombat:


technology/cassandra/src/main/java/com/github/mizool/technology/cassandra/ClusterProducer.java
```diff
@@ -41,6 +43,7 @@
     private static final String
         CASSANDRA_MAX_REQUESTS_PER_CONNECTION_REMOTE_PROPERTY_NAME
         = "cassandra.maxRequestsPerConnection.remote";
+    private static final String CASSANDRA_SOCKET_OPTIONS_READ_TIMEOUT = "cassandra.readTimeoutMillis";
```
bannmann: for consistency with the other constants, add the `_PROPERTY_NAME` suffix

technology/cassandra/src/main/java/com/github/mizool/technology/cassandra/ClusterProducer.java
```diff
@@ -86,6 +90,18 @@ private void registerJdk8TimeCodecs(Cluster cluster)
             .register(LocalTimeCodec.instance);
     }
 
+    private SocketOptions getSocketOptions()
```
bannmann: putting `getSocketOptions()` here is arbitrary, as is the order of the other private methods. reorder them so their order equals the order of calls in `produce()`.

PR #90, opened 2018-07-04
added EnumConverter

voons:


core/src/main/java/com/github/mizool/core/converter/EnumConverter.java
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.mizool.core.converter;
+
+public class EnumConverter
```
bannmann: for consistency with other converters, both methods must be null safe.

core/src/main/java/com/github/mizool/core/converter/EnumConverter.java
```diff
@@ -0,0 +1,30 @@
...
+{
+    public String fromPojo(Enum<?> enumValue)
+    {
+        return enumValue.name();
+    }
+
+    public <T extends Enum<T>> T toPojo(Class<T> enumClass, String value)
+    {
+        return T.valueOf(enumClass, value);
```
bannmann: Won't this cause a compiler warning? maybe the Mizool maven build does not show them yet, but I think the IDE would.

Anyway, I would prefer `Enum.valueOf(enumClass, value);` - it will return a `T` as well.

core/src/main/java/com/github/mizool/core/converter/EnumConverter.java
```diff
@@ -20,11 +20,22 @@
...
+            result = enumValue.name();
+        }
+        return result;
     }
 
     public <T extends Enum<T>> T toPojo(Class<T> enumClass, String value)
     {
-        return T.valueOf(enumClass, value);
+        T result = null;
+        if (enumClass != null && value != null)
```
bannmann: `value` can indeed be null (e.g. in an optional field), that part is correct.

however, if `enumClass` is null, that is a programming error we should not swallow silently. remove that condition, let's run into the NPE thrown by `Enum.valueOf()` in that case.

core/src/main/java/com/github/mizool/core/converter/EnumConverter.java
```diff
@@ -20,11 +20,22 @@
...
+        String result = null;
+
+        if (enumValue != null)
+        {
+            result = enumValue.name();
+        }
+        return result;
     }
 
     public <T extends Enum<T>> T toPojo(Class<T> enumClass, String value)
```
bannmann: for consistency with `fromPojo()`, rename `value` to `enumValue`

PR #91, opened 2018-07-10
Every

JanDoerrenhaus:


core/src/main/java/com/github/mizool/core/Every.java
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Ensures an instance is only ever iterated once to avoid performance issues and memory leaks.
+ */
+public final class Every<T>
+{
+    private final ImmutableSet<T> instances;
+
+    @Inject
+    protected Every(Instance<T> instance)
```
bannmann: Come to think of it, will this actually work? In our other code bases, beans that have an `Instance` injection point usually (always?) specify the type parameter for `Instance`. This one uses `T` instead as it depends on another injection point - the one in the calling class. I fear CDI and type erasure might cause trouble here. Using Google and Stack Overflow, I could not verify or falsify this theory yet.

Please verify that this works, or find an example of a class with a generic type parameter `T` which has an `Instance<T>` injection point.

PR #92, opened 2018-07-11
utility converter for lists

voons:


core/src/main/java/com/github/mizool/core/converter/DtoConverters.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.github.mizool.core.GuavaCollectors;
+
+@UtilityClass
+public class DtoConverters
+{
+    public <D, P> List<D> fromPojos(List<P> pojos, Function<P, D> converter)
+    {
+        List<D> dtos = null;
+        if (pojos != null)
```
bannmann: make this symmetrical to the `toPojos()` method: also return `null` dto for empty pojo list

core/src/main/java/com/github/mizool/core/converter/DtoConverters.java
```diff
@@ -0,0 +1,53 @@
...
+        if (dtos != null)
+        {
+            pojos = dtos.stream().map(converter).collect(GuavaCollectors.toImmutableList());
+        }
+        else
+        {
+            pojos = Collections.emptyList();
+        }
+        return pojos;
+    }
```
bannmann: add overloads for `Map` and `Set`.

core/src/main/java/com/github/mizool/core/converter/DtoConverters.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.function.Function;
+
+import lombok.experimental.UtilityClass;
+
+import com.github.mizool.core.GuavaCollectors;
+
+@UtilityClass
+public class DtoConverters
```
bannmann: We discussed renaming this to `CollectionConverters`, but I'm no longer sure whether that name is a good fit. Our naming scheme seems to imply such a class offers utility methods for classes that are called "collection converter", but we have no such thing. In fact, this class itself comes close: we might turn it into an instantiable class and call it `CollectionConverter`. Then again, it would be an unusual converter as its methods require passing the "element converter". Also, instantiating such a CollectionConverter _and_ the converter for the elements would probably _add_ boilerplate at the call sites, and the goal was to _reduce_ it.

So I tend towards leaving it a utility class, but the name is weird.
@voons @JanDoerrenhaus @schroenser: Any opinions?

Note: the most elegant solution would probably be to introduce a `Converter` interface which adds default methods for `List`/`Set`/`Map` to each converter that implements the interface. But I think this is way out of scope now.

JanDoerrenhaus: @bannmann Maybe just `Converters`?

schroenser: I can't think straight, but `Converters` sounds good, for the time being.

core/src/main/java/com/github/mizool/core/converter/Converters.java
```diff
@@ -0,0 +1,86 @@
...
+    public <D, P> List<D> fromPojos(List<P> pojos, Function<P, D> converter)
+    {
+        List<D> dtos;
+        if (pojos != null)
+        {
+            dtos = pojos.stream().map(converter).collect(GuavaCollectors.toImmutableList());
+        }
+        else
+        {
+            dtos = Collections.emptyList();
```
bannmann: this is not symmetrical behavior. `toPojos(null)` returns an empty list, so `fromPojos(empty)` must return `null`. here's my previous comment:

> make this symmetrical to the `toPojos()` method: also return `null` dto for empty pojo list

do this for the `Set` version as well.

PR #96, opened 2018-07-24
Added utility class for Streams.

schroenser:
@bannmann I just came up with this idea to reduce the spam when streaming iterables. What do you think? Should we add this to mizool and try to use it in the future (or even change existing code)?


core/src/main/java/com/github/mizool/core/Streams.java
```diff
@@ -0,0 +1,20 @@
+package com.github.mizool.core;
```
bannmann: license boilerplate is missing

PR #97, opened 2018-07-25
Feature/shiro mockito extensions

bannmann:
- Added mockSubject() method for Shiro.
- Added BOM to simplify dependency management for downstream projects.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,8 +69,8 @@
     </prerequisites>
 
     <modules>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -70,6 +70,7 @@
 
     <modules>
         <module>core</module>
+        <module>bom</module>
```
schroenser: Sort POMs.

PR #105, opened 2018-08-14
I got quite sick of writing the boilerplate ifs in converters.

JanDoerrenhaus:


core/src/main/java/com/github/mizool/core/converter/IdentifierConverter.java
```diff
@@ -0,0 +1,47 @@
...
+package com.github.mizool.core.converter;
+
+import com.github.mizool.core.Identifiable;
+import com.github.mizool.core.Identifier;
+
+public class IdentifierConverter
+{
+    public String fromPojo(Identifier<?> pojo)
+    {
+        String record = null;
```
schroenser: `com.github.mizool.core.converter` now mixes classes that talk about `result`, `record` and `dto`.

We should think about either breaking it up into sub packages for store and rest layers or try to use the term `result` everywhere, as long as the respective helper class can be used for both layers.

Any thoughts, @bannmann ?

schroenser:
Also we have `Converters`, which is a utility class that helps in dealing with lists and sets and `EnumConverter`, `IdentifierConverter` and `ZonedDateTimeConverter` which follow the pattern of nestable converters.
Thats's not wrong in itself but got me confused at first. I was wondering why this `IdentifierConverter` was not implemented as a utility class (because `Converters` was the first class I opened when I was looking for a reference).

core/src/main/java/com/github/mizool/core/converter/IdentifierConverter.java
bannmann: Let's rename all references to dto/record in this package to `result`. Some classes here are indeed only used in certain layers, but they should not care about that until we find a real clash in either naming or return/argument types.

bannmann: oh, and all method parameters in this package should be named in a neutral way as well, e.g. `String value` instead of `String record` or `String dto`. Referencing "pojo" is of course okay.

core/src/main/java/com/github/mizool/core/converter/IdentifierConverter.java
```diff
@@ -0,0 +1,47 @@
+/**
+ * Copyright 2017-2018 incub8 Software Labs GmbH
+ * Copyright 2017-2018 protel Hotelsoftware GmbH
```
bannmann: as this is a new class, only list `2018`.

PR #111, opened 2018-10-12
Removed Joda-Time dependencies and converted related code to java.time.

schroenser:


core/src/main/java/com/github/mizool/core/validation/CheckPast.java
```diff
@@ -35,19 +34,19 @@ public void initialize(Past past)
...
 
-    private boolean isValidValue(DateTime validationObject)
+    private boolean isValidValue(ZonedDateTime validationObject)
     {
         return validationObject.isBefore(now());
     }
 
     @VisibleForTesting
-    protected DateTime now()
+    protected ZonedDateTime now()
```
bannmann: let's simplify the code and unit test by using a `Clock` field instead:

1. add a no-args constructor to that sets the field to [`Clock.systemDefaultZone()`](https://docs.oracle.com/javase/8/docs/api/java/time/Clock.html#systemDefaultZone--).
2. the unit test passes [`Clock.fixed(...)`](https://docs.oracle.com/javase/8/docs/api/java/time/Clock.html#fixed-java.time.Instant-java.time.ZoneId-) to a `@VisibleForTesting CheckPast(Clock)` constructor.

PR #115, opened 2018-11-28
the EnumConverter  is now able to handle lists of enumeration values

codingWombat:


schroenser:
I'm only the reviewer when @bannmann opened the pull request.

core/src/main/java/com/github/mizool/core/converter/EnumConverter.java
```diff
@@ -40,4 +45,26 @@ public String fromPojo(Enum<?> enumValue)
         }
         return result;
     }
+
+    public List<String> fromPojos(List<Enum<?>> enumValues)
```
bannmann: this is redundant with `com.github.mizool.core.converter.Converters` introduced by PR #92. also, it does not comply with OMG-27 in that it passes `null` collections to the core layer.

core/src/main/java/com/github/mizool/core/validation/CheckEnumValue.java
```diff
@@ -16,10 +16,12 @@
  */
 package com.github.mizool.core.validation;
 
+import java.util.List;
+
 import javax.validation.ConstraintValidator;
 import javax.validation.ConstraintValidatorContext;
 
-public class CheckEnumValue implements ConstraintValidator<EnumValue, String>
+public class CheckEnumValue implements ConstraintValidator<EnumValue, Object>
```
bannmann: before we adapt this check (and potentially others), we should try using `@EnumValue(...) List<String>` (BeanVal < 2.0) and `List<@EnumValue(...) String>` (as of BeanVal 2.0, using a JDK 8 language change allowing annotations on type usages and not only declarations).

core/src/main/java/com/github/mizool/core/converter/EnumConverter.java
bannmann: see also https://beanvalidation.org/proposals/BVAL-508/

core/src/main/java/com/github/mizool/core/validation/CheckEnumValue.java
bannmann: see also https://beanvalidation.org/proposals/BVAL-508/

core/src/main/java/com/github/mizool/core/converter/EnumConverter.java
schroenser: The calls to use the `Converters` class are
`Converters.fromPojos(foos, enumConverter::fromPojo)`
and
`Converters.toPojos(fooDtos, fooDto -> enumConverter.toPojo(Foo.class, fooDto))`
respectively.

JanDoerrenhaus:
@codingWombat Please update this Pull Request, it ages badly.

codingWombat:
as @schroenser statet thare are already calls that cover this scenario

PR #118, opened 2019-01-10
added new exception

voons:
After a short discussion we came up with this new exception. It might be interesting for your purposes as well @schroenser.


bannmann:
fyi @JanDoerrenhaus

PR #119, opened 2019-01-18
Improved ease of use when waiting on a semaphore for a condition. Rem…

schroenser:
…oved some SuppressWarnings on squids in the process.


voons:
fyi @JanDoerrenhaus

PR #123, opened 2019-04-09
revisiting 'iterable' support in constraint validators

bannmann:
up for discussion, do not merge yet

### DIFF
--- a/core/src/test/java/com/github/mizool/core/validation/TestConstraintValidators.java
+++ b/core/src/test/java/com/github/mizool/core/validation/TestConstraintValidators.java
@@ -66,25 +66,54 @@ public class TestConstraintValidators
             // a null object is perfectly fine for optional fields
             { "null as value of an optional field", Type.OPTIONAL, null, PREDICATE_NOT_CALLED, true },
 
+            // we decided that mandatory = true" accepting an empty list would not be intuitive
+            { "empty list in mandatory field", Type.MANDATORY, Collections.emptyList(), PREDICATE_NOT_CALLED, false },
+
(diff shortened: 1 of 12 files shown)


core/src/test/java/com/github/mizool/core/validation/TestConstraintValidators.java
```diff
@@ -66,25 +66,54 @@ public void test(
...
+
+    other requirements:
+        - when a field List<Integer> is annotated with a beanval check that calls ConstraintValidators with
+          a predicate for T=String, this is a CodeInconsistencyException.
+
+
+
+    notes:
+     - to satisfy jens' curiosity, test this: when a DTO field wants a list of strings, but the client-supplied JSON
+       has a list of numbers, we should never even reach a bean validation check.
```
voons: sadly we do, the postDto in the resource contained a list of Strings ["1","4","5"] instead of [1,4,5].

bannmann: okay, so the json mapping code (jackson/gson/whatever) is lenient and will convert values to the type specified in the DTO class if possible (like in the case mentioned by @voons). while it makes our APIs less strict than I would like them to be, I think this does not create immediate problems for us and can be ignored for now.

it would be more interesting to see what happens for the reverse case: a DTO with an `Integer` or `List<Integer>` field and the client passes JSON values like `"hello"` and `["hello", "world"]`. I would expect an exception. in that case, we don't need to do anything about that, but it is valuable background information when discussing type safety and/or (lack of) generics in our validator classes.

schroenser:
We decided to close this, as it's not worth investigating what if anything of this is still relevant.

PR #124, opened 2019-05-15
Feature/add dto capabilities to converters utility class

schroenser:
I've performed this change as I've seen an increasing amount of `null`-safe `Collection` conversion across several converters.

We do _not_ want to introduce any side effects regarding `null` vs. empty collections in the REST layer but I still think it would greatly reduce visual clutter from the converters if we were able to use this utility class in both REST and store layers.


bannmann:
fyi @JanDoerrenhaus

PR #126, opened 2019-05-23
Replaced white list mechanism with exception behaviours.

JanDoerrenhaus:


core/src/main/java/com/github/mizool/core/rest/errorhandling/behavior/InvalidBackendReplyExceptionBehavior.java
```diff
@@ -0,0 +1,49 @@
...
+    @Override
+    public boolean includeErrorId()
+    {
+        return true;
+    }
+
+    @Override
+    public boolean includeDetails()
+    {
+        return true;
```
schroenser: This has changed with regards to the old `DefaultWhiteList`. Was that intentional or copy and paste?

JanDoerrenhaus: This was intentional, as the old behaviour was one of the motivations behind this change. Just to control the status code, we whitelisted this exception, resulting in us being blind in cases of backend to backend communication breaking down.

schroenser:
fyi: @bannmann

PR #130, opened 2019-07-01
Added support for java.util.Optional to Optionals utility class.

schroenser:
There's no particular reason why I did this now. I was just thinking about it and felt like doing it right away before writing a ticket or the like.

The only downside we might get from this is the huge amount of deprecation warnings until we manage to adapt our other projects.


bannmann:
fyi @JanDoerrenhaus

PR #131, opened 2019-07-01
Added Optionals.mapToValue() to simplify stream pipelines.

bannmann:
I had this in a private codebase in a `MoreOptionals` helper class, but IMO it really belongs here as other codebases can benefit from it. 

Method name is up for debate.

### DIFF
--- a/core/src/main/java/com/github/mizool/core/Optionals.java
+++ b/core/src/main/java/com/github/mizool/core/Optionals.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Copyright 2017-2019 incub8 Software Labs GmbH
  * Copyright 2017-2019 protel Hotelsoftware GmbH
  *
@@ -119,7 +119,7 @@ public class Optionals
      *     .map(Optional::get)
      * }</pre>
      */
-    public <T> Stream<T> streamPresentValue(Optional<T> optional)
+    public <T> Stream<T> streamPresentValue(@NonNull Optional<T> optional)
     {
         return optional.map(Stream::of).orElseGet(Stream::empty);
     }
--- a/core/src/main/java/com/github/mizool/core/Optionals.java
+++ b/core/src/main/java/com/github/mizool/core/Optionals.java
@@ -106,11 +106,12 @@ public class Optionals
     }
 
     /**
-     * Used in streams to {@linkplain Stream#flatMap(Function) flat-map} each {@link Optional} to its value if present.<br>
+     * Used in streams to {@linkplain Stream#flatMap(Function) flat-map} each {@link Optional} to its value if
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/github/mizool/core/TestOptionals.java
```diff
@@ -0,0 +1,59 @@
...
+        Optional<String> b = Optional.of("b");
+        Optional<String> c = Optional.of("c");
+
+        return new Object[][]{
+            { emptyList, emptyList },
+            { ImmutableList.of(empty), emptyList },
+            { ImmutableList.of(a), ImmutableList.of("a") },
+            { ImmutableList.of(a, b, c), ImmutableList.of("a", "b", "c") },
+            { ImmutableList.of(a, empty, c), ImmutableList.of("a", "c") }
+        };
```
schroenser: Add `{ "multiple equal values", ImmutableList.of(a, a), ImmutableList.of("a", "a") }` to ensure there is no reduction.

core/src/test/java/com/github/mizool/core/TestOptionals.java
```diff
@@ -0,0 +1,59 @@
...
+
+import com.google.common.collect.ImmutableList;
+
+public class TestOptionals
+{
+    @Test(dataProvider = "optionalStreams")
+    public <T> void testMapToValue(Collection<Optional<T>> input, Iterable<T> expected)
+    {
+        List<T> actual = input.stream().flatMap(Optionals::mapToValue).collect(Collectors.toList());
+        assertThat(actual).containsExactlyElementsOf(expected);
```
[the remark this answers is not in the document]
bannmann: Good point!

Also, to prevent readers from thinking "ah, it's a list, so order is relevant", I switched from `List` to `Set`. This is just a hint, though, because `ImmutableSet` also makes guarantees about iteration order.

core/src/main/java/com/github/mizool/core/Optionals.java
```diff
@@ -106,19 +106,20 @@
...
+     *     .flatMap(Optionals::streamPresentValue)
      * }</pre>
      * Using this method is equivalent of chaining {@link Optional#isPresent()} and {@link Optional#get()} like this:
      * <pre>{@code
      *     .filter(Optional::isPresent)
      *     .map(Optional::get)
      * }</pre>
      */
-    public <T> Stream<T> mapToValue(Optional<T> optional)
+    public <T> Stream<T> streamPresentValue(@NonNull Optional<T> optional)
```
schroenser: Good catch on the `@NonNull`.

schroenser:
fyi: @JanDoerrenhaus

PR #132, opened 2019-07-21
Dto validation.

JanDoerrenhaus:


core/src/main/java/com/github/mizool/core/validation/dto/CheckPostDto.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private boolean isValidValue(Dto dto)
+    {
+        boolean valid = false;
+        if (dto.getId() == null &&
+            dto.getTimestamp() == null &&
+            dto.getCreationTimestamp() == null &&
+            dto.getModificationTimestamp() == null)
+        {
+            valid = true;
```
schroenser: `if something then result = true` smells like poodoo. Change the method to `return dto.getId() == null && dto.getTimestamp() == null && dto.getCreationTimestamp() == null && dto.getModificationTimestamp() == null;`.

core/src/main/java/com/github/mizool/core/validation/dto/CheckPutDto.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private boolean isValidValue(Dto dto)
+    {
+        boolean valid = false;
+        if (dto.getId() != null &&
+            (dto.getTimestamp() != null ||
+                (dto.getCreationTimestamp() != null &&
+                dto.getModificationTimestamp() != null)))
+        {
+            valid = true;
```
schroenser: `if something then result = true` smells like poodoo. Change the method to `dto.getId() != null && (dto.getTimestamp() != null ^ (dto.getCreationTimestamp() != null && dto.getModificationTimestamp() != null));`.
Also note the `^` (XOR) operator, as we also want to make sure that not all three timestamps are set simultaneously.

core/src/main/java/com/github/mizool/core/validation/dto/PostDto.java
```diff
@@ -0,0 +1,41 @@
...
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import javax.validation.Constraint;
+import javax.validation.Payload;
+
+@Target({ FIELD, PARAMETER })
```
schroenser: Note that `ElementType.FIELD` is allowed to support injecting dtos into fields. I was first confused as this _also_ allows to annotate sub-dtos inside other dtos, which doesn't make much sense.

core/src/main/java/com/github/mizool/core/validation/dto/PutDto.java
```diff
@@ -0,0 +1,41 @@
...
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import javax.validation.Constraint;
+import javax.validation.Payload;
+
+@Target({ FIELD, PARAMETER })
```
schroenser: Note that `ElementType.FIELD` is allowed to support injecting dtos into fields. I was first confused as this _also_ allows to annotate sub-dtos inside other dtos, which doesn't make much sense.

PR #133, opened 2019-07-22
Configuration that is implemented by system properties in the CDI wor…

JanDoerrenhaus:
…ld, or by environment variables in the Guice world.


core/src/main/java/com/github/mizool/core/configuration/Configuration.java
```diff
@@ -0,0 +1,8 @@
+package com.github.mizool.core.configuration;
+
+public interface Configuration
+{
+    String get(String key);
+
+    String get(String key, String def);
```
schroenser: Rename `def` to `default`. I think `System.getProperty(String key, String def)` was defined back in the day where the poor developers had to actually press keys for every single letter in their variable names and/or were not payed for "bytes of code produced".

core/src/main/java/com/github/mizool/core/configuration/EnvironmentVariableConfigurationImpl.java
```diff
@@ -0,0 +1,26 @@
...
+    public String get(String key)
+    {
+        return System.getenv(key);
+    }
+
+    @Override
+    public String get(String key, String def)
+    {
+        String result = System.getenv(key);
+        if (Strings.isNullOrEmpty(result))
```
schroenser: This is not syntactically correct.
`System.getProperty(String key, String def)` will _only_ return the default value when the system property is not defined. If it's defined but empty, an empty string is returned.
Since `System.getenv(String name)` will also return an empty string if the environment variable is defined but empty and only return `null` when the environment variable is not defined, we have to check for `result == null` to imitate the behavior of `System.getProperty(String key, String def)`.

core/src/main/java/com/github/mizool/core/configuration/ConfigurationModule.java
```diff
@@ -0,0 +1,17 @@
+package com.github.mizool.core.configuration;
+
+import org.kohsuke.MetaInfServices;
+
+import com.google.inject.AbstractModule;
+import com.google.inject.Module;
+
+@MetaInfServices(Module.class)
+public class ConfigurationModule extends AbstractModule
```
schroenser: This is the wrong place for this module. Loading environment variables instead of system properties came up with lambdas. Putting it here as a service provider for the `com.google.inject.Module` interface hardwires guice to to environment variables and thus lambdas (at least for now).

The opposite goes for the `@Vetoed` annotation on the `EnvironmentVariableConfigurationImpl`.

I wonder if `ConfigurationModule` and `EnvironmentVariableConfigurationImpl` belong into `technology/lambda`. However, I don't see where to put `SystemPropertyConfigurationImpl` so it doesn't collide. `technology/not-lambda` sounds cumbersome.

We could put `ConfigurationModule` and `EnvironmentVariableConfigurationImpl` into a lambda technology and have `EnvironmentVariableConfigurationImpl` extend `SystemPropertyConfigurationImpl` with a `@Specializes` annotation. But that feels wrong as well.

Perhaps @bannmann has an idea when he reviews this?

JanDoerrenhaus: My suggestion would be to move this to technology-lambda once we actually create that technology when the lambda stuff from genius has matured.

JanDoerrenhaus: To be more precise: The vetoed implementation offers an alternative that is inactive by default. Any consumer is free to activate it in any way deemed proper. Once the module that activates it moves to the lambda technology, the concept becomes clean again.

bannmann: Let's solve this dilemma as follows:

- `SystemPropertyConfigurationImpl` is kept as is. CDI based applications will use it by default.
- To avoid a general "guice means environment variable config" assumption in mizool-core, `ConfigurationModule` and `EnvironmentVariableConfigurationImpl` remain in the downstream project (Genius) for now. They keep their respective `@MetaInfServices` and `@Vetoed` annotations. 
    - When we move the lambda stuff from downstream to mizool, `ConfigurationModule` ends up in `technology-lambda`. At that point, we can move `EnvironmentVariableConfigurationImpl` there as well or to core, but neither option should cause problems for the CDI world.

PR #140, opened 2019-07-24
Added support for LocalDateTime as record typ to ZonedDateTimeConverter. GENIUS-1531

schroenser:
GENIUS will access its database using jOOQ. jOOQ natively supports `LocalDateTime` for `timestamp` fields, so I added support for that to the `ZonedDateTimeConverter`.


bannmann:
FYI @JanDoerrenhaus

PR #141, opened 2019-07-25
Added store converter to serialize complex pojo fields into JSON. GENIUS-1531

schroenser:
In genius, we will use this to store sets and maps without having to normalize our tables (which would complicate things in the store layer). The idea to use JSON serialization for that is borrowed from the Cassandra world, where we had problems using Cassandras collection support and decided to perform the serialization ourselves. Sadly, since this decision was made, there was no time to implement it anywhere.

Note that since this converter can't possibly traverse the object tree to check for `null` collections, the converter that calls this one will have to wrap the call if needed.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -18,6 +18,11 @@
             <artifactId>jcommander</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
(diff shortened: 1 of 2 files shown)


schroenser:
Nasser told me that there might be problems with more complex objects. I'm closing this pull request and will add a test class that demonstrates what works.

PR #142, opened 2019-07-30
Added store converter to serialize complex pojo fields into JSON. GENIUS-1531

schroenser:


core/src/main/java/com/github/mizool/core/converter/JsonConverter.java
```diff
@@ -0,0 +1,53 @@
...
+            }
+            catch (JsonProcessingException e)
+            {
+                throw new StoreLayerException("Error serializing field", e);
+            }
+        }
+        return record;
+    }
+
+    public <P> P toPojo(String record, TypeReference<P> type)
```
bannmann: I have the feeling that this API is not optimal and makes the call sites ugly.

Can we maybe replace the `TypeReference<P>` with `Class<P>`?

Either way, I think that having this typing as a type parameter on the class level (together with a constructor argument) would feel better (even though `toRecord(P pojo)` does not require it). Doing that would be make this class a bit more similar to regular converters, which are strictly tied to one pair of `P, R` (even though those are usually not generic). For example, imagine we had an interface for `Converter<P, R>`, then we could have `JsonConverter<P> implements Converter<P, String>`.

schroenser: No we can't. If you look at `TestJsonConverter.testCollectionToPojo` and `TestJsonConverter.testMapToPojo`, you can see that the `TypeReference` allows to pass types that have type parameters themselves. If the parameter was `Class<P>`, I could only pass `Collection.class` or `Map.class` respectively, making it impossible for the `ObjectMapper` to do the right thing.

The type parameter can also not be moved to class level, as a constructor parameter would force the developer to create `@Produces` methods for every concrete `TypeReference<P>` they want to use.
If for example my injection point in another converter is `JsonConverter<Map<String, String>>`, there is no way the injector can satisfy the the `TypeReference<Map<String, String>>` parameter of the constructor. Unless I provide a
```java
@Produces
protected TypeReference<Map<String, String>> produce()
{
    return new TypeReference<Map<String, String>>()
    {
    };
}
```
The use case for this converter is to allow for a pojo converter do decide to have one of the pojo's fields to be serialized and deserialized to json. Having the parameter on the method keeps the knowledge about the pojo's field type and how to map it ocontained inside the pojo converter, where it belongs.

PR #143, opened 2019-08-01
added mapper option to write unset instead of null into columns, prev…

voons:
we recently discovered that cassandra creates a tombstone for each cell when value of the cell is null.
See this article for details: https://opencredo.com/blogs/cassandra-tombstones-common-issues/
To prevent this behvior, the mapper has an option to store a cell with value unset instead of null.

### DIFF
--- a/technology/cassandra/src/main/resources/com/github/mizool/technology/cassandra/mapperProducerGenerator.stg
+++ b/technology/cassandra/src/main/resources/com/github/mizool/technology/cassandra/mapperProducerGenerator.stg
@@ -45,6 +45,7 @@ class $className$MapperProducer
         mapper.setDefaultGetOptions(Option.consistencyLevel(DefaultConsistencyLevel.READ));
         mapper.setDefaultSaveOptions(Option.consistencyLevel(DefaultConsistencyLevel.WRITE));
         mapper.setDefaultDeleteOptions(Option.consistencyLevel(DefaultConsistencyLevel.WRITE));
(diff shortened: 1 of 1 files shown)


bannmann:
This will break when `null` is written to a cell that previously contained a non-`null` value.

Example:

> Node 1: 'a=42 timestamp=2016-01-01T13:37:00.000'
> Node 2: 'a=42 timestamp=2016-01-01T13:37:00.000'
> Node 3: 'a=42 timestamp=2016-01-01T13:37:00.000'

Now we write an entity that has `a=null` using the mapper, and two nodes succeed:

> Node 1: 'a=42 timestamp=2016-01-01T13:37:00.000'
> Node 2: 'TOMBSTONE timestamp=2019-08-05T12:46:00.000'
> Node 3: 'TOMBSTONE timestamp=2019-08-05T12:46:00.000'

Node 1 did not get the write. But nothing bad happens: when the cell is read (with quorum consistency), the coordinator node for this query will notice the conflicting information, use the timestamp information to select a winner (the deletion) and issue a read repair for Node 1. In any case, before the tombstones disappear, our scheduled repair job will use the timestamp information to fix it the inconsistency.

If we had set `saveNullFields(false)` for the write, the result would look like this:

> Node 1: 'a=42 timestamp=2016-01-01T13:37:00.000'
> Node 2: _no information_
> Node 3: '_no information_

Both a read repair and the scheduled repair would then propagate the info from Node 1, resurrecting the cell and maybe even causing the zombie apocalypse.

PR #148, opened 2019-10-02
Added StoreLayerFuture

lesmou:


bannmann:
FYI @JanDoerrenhaus

technology/cassandra/src/main/java/com/github/mizool/technology/cassandra/StoreLayerFuture.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @Override
+    public Void get() throws InterruptedException, ExecutionException
+    {
+        try
+        {
+            return target.get();
+        }
+        catch (QueryExecutionException | QueryValidationException | NoHostAvailableException e)
```
JanDoerrenhaus: Did't we say that the QueryValidationException should lead to a CodeInconsistencyException rather than a StoreLayerException?

bannmann: Right, we did, and the classes in the downstream codebase which use synchronous statements were adapted accordingly. But this code was forgotten.

@lesmou, please create a new PR fixing this.

PR #151, opened 2019-11-18
include details in LockedEntityException

voons:
we would like to include details in locked entity exceptions for frontends to distinguish between different kind of locks


voons:
I already discussed this with @schroenser and he sees no problem with including details.

bannmann:
FYI @JanDoerrenhaus

PR #153, opened 2019-12-19
Feature/health check framework

schroenser:


pom.xml
```diff
@@ -96,6 +96,7 @@
         <java.version>1.8</java.version>
         <jcache.ri.version>0.5</jcache.ri.version>
         <jersey.version>2.25</jersey.version>
+        <jooq.version>3.11.11</jooq.version>
```
lesmou: Maybe think about using the most recent version of [jOOQ](https://www.jooq.org/notes) which is [3.12.3](https://www.jooq.org/notes#3.12.3) since end of October.

technology/web/src/main/java/com/github/mizool/technology/web/healthcheck/HealthCheck.java
```diff
@@ -0,0 +1,66 @@
...
+    private Set<Check> asSetOrDefault(Check[] checks)
+    {
+        Set<Check> result;
+        if (checks.length == 0)
+        {
+            result = ImmutableSet.of(new DefaultCheck());
+        }
+        else
+        {
+            result = ImmutableSet.copyOf(checks);
```
lesmou: Do we also want the default check to be always run even if specialised checks are given?

schroenser: I'd vote against that as the default check does nothing but produce an "ok" when asked. So if there are actual checks present, having an additional `"default": "ok"` in the response does not add value.

lesmou: I was thinking of the `default` in addition to make clear, that we did not only check the database connectivity but also the application itself, even though this is implicit since the was a response... d'uh...
I just wanted to have a complete list of "working" system parts in the response rather than a list of "some" of the parts we checked and the implicit information that the application itself is fine as well.
I don't insist so maybe we leave it as is and see if there are more opinions on that.
@bannmann

bannmann: > I just wanted to have a complete list of "working" system parts

Good thinking, and I agree with the goal. However, as `DefaultCheck` does not really represent a _system part_, I'm with @schroenser on this.

Put another way: if an application _did_ check more than the database connectivity, it will add a `Check` implementation for that, and if it doesn't, it's pretty accurate that the health check resource does not say anything about "the rest of the application".

Outlook: maybe in the future we will provide more standard `Checks` in Mizool which applications can easily reuse, or maybe those checks are even always added (and at that point we would likely drop the `DefaultCheck`). We'll see.

technology/web/src/main/java/com/github/mizool/technology/web/healthcheck/Check.java
```diff
@@ -0,0 +1,6 @@
+package com.github.mizool.technology.web.healthcheck;
```
bannmann: add license boilerplate comments to all new java files

technology/web/src/main/java/com/github/mizool/technology/web/healthcheck/JooqConnectivityCheck.java
```diff
@@ -0,0 +1,36 @@
...
+        CheckResult.CheckResultBuilder resultBuilder = CheckResult.builder().name(name);
+
+        try (DSLContext dslContext = dslContextSupplier.get())
+        {
+            dslContext.selectOne().execute();
+            resultBuilder = resultBuilder.success(true).message("OK");
+        }
+        catch (DataAccessException e)
+        {
+            log.debug("Error checking the connecting to " + name, e);
```
bannmann: typo: `connection`

technology/web/src/main/java/com/github/mizool/technology/web/healthcheck/Check.java
schroenser: Can we _please_ change the license already? *growls*

If I knew how, I would have already done that.

PR #157, opened 2020-01-08
Feature/add sql technology

schroenser:
Review after pr #156 was merged.

### DIFF
--- a/technology/sql/src/main/java/com/github/mizool/technology/sql/jooq/MizoolConnectionProvider.java
+++ b/technology/sql/src/main/java/com/github/mizool/technology/sql/jooq/MizoolConnectionProvider.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2020 incub8 Software Labs GmbH
+ * Copyright 2020 protel Hotelsoftware GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.mizool.technology.sql.jooq;
+
+import lombok.Getter;
+
+import org.jooq.ConnectionProvider;
+import org.jooq.SQLDialect;
+
+import com.github.mizool.core.exception.CodeInconsistencyException;
+
+public abstract class MizoolConnectionProvider implements ConnectionProvider
+{
+    @Getter
+    private final SQLDialect sqlDialect;
+
+    protected MizoolConnectionProvider(String driverClassName, SQLDialect sqlDialect)
+    {
+        this.sqlDialect = sqlDialect;
+        System.setProperty("org.jooq.no-logo", "true");
+        loadDriver(driverClassName);
+    }
+
+    private void loadDriver(String driverClassName)
+    {
+        try
+        {
+            Class.forName(driverClassName);
+        }
+        catch (ClassNotFoundException e)
+        {
+            throw new CodeInconsistencyException("JDBC driver '" + driverClassName + "' is missing on classpath", e);
+        }
+    }
+}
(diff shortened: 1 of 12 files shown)


technology/sql/src/main/java/com/github/mizool/technology/sql/jooq/MizoolConnectionProvider.java
```diff
@@ -0,0 +1,49 @@
...
+import org.jooq.SQLDialect;
+
+import com.github.mizool.core.exception.CodeInconsistencyException;
+
+public abstract class MizoolConnectionProvider implements ConnectionProvider
+{
+    @Getter
+    private final SQLDialect sqlDialect;
+
+    protected MizoolConnectionProvider(String driverClassName, SQLDialect sqlDialect)
```
bannmann: add fail-early behavior by adding lombok `@NonNull` to `sqlDialect`. for consistency, add the same to `driverClassName`.

technology/sql/src/main/java/com/github/mizool/technology/sql/jooq/ResultQueries.java
```diff
@@ -0,0 +1,63 @@
...
+import org.jooq.Table;
+import org.jooq.TableRecord;
+import org.jooq.exception.DataAccessException;
+
+import com.github.mizool.core.exception.StoreLayerException;
+
+@UtilityClass
+public class ResultQueries
+{
+    public <R extends TableRecord<R>> Optional<R> fetchOptional(ResultQuery<R> resultQuery, Table<R> table)
```
bannmann: `fetchOptional()` is named like the original JOOQ method, but I find the name to be misleading: it makes these two methods look like you can choose between "an optional" and "a stream" when in fact you are fetching either **a single record** (our classic `read` store operation, and I imagine it would also be used for `findBy...`) or **multiple records** (our classic `list` store operations).

I suggest we rename this to `fetchOptional()` to `read()` and renaming `fetchStream()` to `list()`. Alternatively, one could use something like `fetchSingle()` and `fetchMultiple()`, but I prefer the read/list wording.

technology/sql/src/main/java/com/github/mizool/technology/sql/jooq/ResultQueries.java
```diff
@@ -0,0 +1,63 @@
...
+    {
+        try
+        {
+            /*
+             * Note that we perform .fetch(), so all results are loaded into memory and the query can be closed. The
+             * results are then transformed into a stream for convenience.
+             * The ResultQuery directly supports .stream(). However, this is indeed a lazy operation on a cursor, and
+             * we would have to make sure the stream is properly closed after it has been consumed.
+             */
+            return resultQuery.fetch().stream();
```
bannmann: I'm not convinced by the reasoning given above. What if I **really** want to stream the data because I have to read thousands of rows? This method would be useless to me in that scenario.

The other way round works fine, though: if this method behaved in a lazy fashion (registering a close handler on the returned stream which closes the query), it's up to the caller (the store class): it could keep the streaming aspect (by adding converter calls to the stream chain and returning that stream) or load everything into memory (by collecting the stream into a list and returning either that list or the "safe" stream).

technology/sql/src/main/java/com/github/mizool/technology/sql/jooq/ResultQueries.java
```diff
@@ -0,0 +1,62 @@
...
+        {
+            throw new StoreLayerException("Error fetching optional " + table.getName(), e);
+        }
+    }
+
+    /**
+     * @return A live {@code Stream} of the result data. The underlying connection is still open. You have to make sure
+     * the {@code Stream} is properly closed, either by <b>always</b> consuming it fully (risky due to exceptions) or
+     * preferrably by consuming it inside a <i>try-with-resources</i></i> block.
+     */
```
schroenser: Note that I personally do not like this, since i think that it's easy to forget to _try-with-resources_ a `Stream`.
We usually don't do that and now it's the job of the REST layer, which consumes the `Stream`, to remember to use _try-with-resources_ because of something the store layer does.
Either that, or **all** of our other usages of `Stream`s are wrong, because we _always_ should use _try-with-resources_ as we can never know if it's crucial that a particular `Stream` is closed.
Thoughts?

bannmann: As much as it hurts, we have to do the second approach: if you get a stream, you need to use try-with-resources.

bannmann:
FYI @JanDoerrenhaus

PR #160, opened 2020-01-27
shiro-web is no longer optional for applications depending on the web…

schroenser:
…-technology.


bannmann:
FYI @JanDoerrenhaus

PR #163, opened 2020-02-28
Add HttpFilterAdapter

bannmann:


schroenser:
Please also have a look at the issues in Sonar, that are in the changed classes. Sadly our Travis builds currently are not correctly connected to the branches, so you will have to search the list of all issues for this class. I've seen that there are a couple of problems.

schroenser:
fyi @JanDoerrenhaus

PR #164, opened 2020-02-28
added argon2 hasher

voons:


core/src/main/java/com/github/mizool/core/password/Argon2Hasher.java
```diff
@@ -0,0 +1,56 @@
...
+    public String hashPassword(char[] plaintextPassword)
+    {
+        String result;
+        try
+        {
+            result = argon2.hash(ITERATIONS_FOR_NEW_PASSWORDS, MEMORY, PARALLELISM, plaintextPassword);
+        }
+        finally
+        {
+            argon2.wipeArray(plaintextPassword);
```
bannmann: Wiping arrays that contain credentials is important, but in this case it doesn't belong here: this particular array was supplied by the calling code, so that code must be the one that wipes it (not via argon2 of course). If the hasher did it, it would be an undocumented side effect.

Please add a Javadoc comment to both `PasswordHasher` methods, adding a note to the parameter description that reminds the reader that the calling code should wipe the array immediately after hashing/verification.

bannmann: oh, and mention "in a try-finally block".

core/src/main/java/com/github/mizool/core/password/Argon2Hasher.java
```diff
@@ -0,0 +1,56 @@
...
+    public boolean passwordsMatch(char[] submittedPlaintext, String digest)
+    {
+        boolean result;
+        try
+        {
+            result = argon2.verify(digest, submittedPlaintext);
+        }
+        finally
+        {
+            argon2.wipeArray(submittedPlaintext);
```
bannmann: see above.

bannmann:
FYI @JanDoerrenhaus

PR #165, opened 2020-04-06
Feature/genius 2034 access private and trusted keys in their normal form

schroenser:
I took the liberty to extract this from the applications so the code remains consistent.


lesmou:
Added @JanDoerrenhaus since @bannmann is on holiday.

core/src/main/java/com/github/mizool/core/rsa/RsaKeys.java
```diff
@@ -0,0 +1,96 @@
...
+            result = (RSAPrivateKey) keyFactory.generatePrivate(keySpec);
+        }
+        catch (InvalidKeySpecException e)
+        {
+            throw new ConfigurationException("Invalid key specification", e);
+        }
+        return result;
+    }
+
+    public RSAPublicKey publicKeyFromX509(String src)
```
JanDoerrenhaus, line 48: Public before private.

core/src/main/java/com/github/mizool/core/rsa/RsaKeys.java
```diff
@@ -0,0 +1,96 @@
...
+            result = (RSAPublicKey) keyFactory.generatePublic(keySpec);
+        }
+        catch (InvalidKeySpecException e)
+        {
+            throw new ConfigurationException("Invalid key specification", e);
+        }
+        return result;
+    }
+
+    private static String stripHeaderFooterAndLineBreaks(String src)
```
JanDoerrenhaus, line 78: Static before non-static.

core/src/main/java/com/github/mizool/core/rsa/RsaKeys.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.regex.Pattern;
+
+import lombok.experimental.UtilityClass;
+
+import com.github.mizool.core.exception.ConfigurationException;
+
+@UtilityClass
+public class RsaKeys
+{
+    private static final Pattern HEADER_FOOTER_AND_LINE_BREAKS = Pattern.compile(
```
JanDoerrenhaus, line 23: Oh, right. This is a @UtilityClass. Then remove the static here as well.

schroenser: I already had the same idea. However, when I removed the keyword, Sonar started to complain that the result of `Pattern.compile` should always be assigned to a `static final` field. Looks like the sonar rule does not know about Lombok.
I'm afraid that is a conflict of interest with [OMG-29](https://incub8.atlassian.net/browse/OMG-29). I'm not sure which route we should take with this.

I think, I'll remove the keyword here, giving our decision in the OMG precedence, we mark this as false positive in Sonar and talk about this during one of our next meetings.

schroenser: I've created [CCCP-34](https://incub8.atlassian.net/browse/CCCP-34) to further this discussion.

Issue #147, opened 2019-09-30
Improve ErrorHandlingCatalog


bannmann:
Extract methods so that the condition
```java
!catalog.containsKey(throwableClass) || !behaviorPackage.startsWith(CORE_PACKAGE)
```
becomes
```java
notInCatalog(throwableClass) || isApplicationLevel(behavior)
```
Furthermore, add debug logging in the `else` branch of that `if`. 

Also, unify `behaviour`/`behavior` to the American English version.

PR #167, opened 2020-04-27
Feature/spring cleanup of error handling

JanDoerrenhaus:


core/src/main/java/com/github/mizool/core/rest/errorhandling/ErrorResponseFactory.java
```diff
@@ -79,14 +66,14 @@ else if (isAssignable(RuleViolationException.class, cursor))
             }
             else if (isAssignable(ClientErrorException.class, cursor))
             {
-                result = errorMapper.handleClientError((ClientErrorException) cursor);
+                result = clientErrorMapper.handleClientError((ClientErrorException) cursor);
```
bannmann, line 69: I wonder why we (even before this PR) `break` if `genericErrorMapper` found something, but don't do the same for the other `isAssignable()` checks. Why do we continue iterating the `cause` chain, overwriting a response `result` we already determined? I can only imagine that this bug just doesn't occur in practice because `ConstraintViolationException`, `RuleViolationException` and `ClientErrorException` are unlikely to even have cause exceptions. Still, we should be consistent here.

Please also add a unit test (for at least one case) that verifies we stop iterating the `cause` chain as soon as we have a non-null behavior or `handle*()` result.

core/src/main/java/com/github/mizool/core/rest/errorhandling/GenericErrorMapper.java
```diff
@@ -31,39 +29,42 @@
...
-        Class<? extends Exception> errorClass = determineErrorClass(statusCode, e.getClass());
-
-        ErrorDto error = new ErrorDto(errorClass.getName(), null);
-        ErrorMessageDto errorMessage = createErrorMessageDto(error);
-        return new ErrorResponse(statusCode, errorMessage);
+        ErrorResponse result = null;
+        Optional<ErrorHandlingBehavior> behaviorOptional = errorHandlingBehaviorCatalog.lookup(t);
+        if (behaviorOptional.isPresent())
+        {
+            ErrorHandlingBehavior behaviour = behaviorOptional.get();
```
bannmann, line 50: `if-isPresent-get` is discouraged and should usually be replaced with `ifPresent()` or `ifPresentOrElse()` (JDK 9). That doesn't work here due to the write to `result`, but we can still get rid of `isPresent-get` by turning this into `ErrorResponse result = behaviorOptional.map(...).orElse(null)` (either with an inline lambda or a method reference).

Also, rename the local variable `behaviour` to `behavior`.

core/src/main/java/com/github/mizool/core/rest/errorhandling/ErrorResponseFactory.java
```diff
@@ -27,28 +24,19 @@
...
-    {
-        errorHandlingCatalog = new ErrorHandlingCatalog();
-        errorMapper = new ErrorMapper();
-        constraintViolationMapper = new ConstraintViolationMapper();
-        ruleViolationMapper = new RuleViolationMapper();
-    }
-
-    @Inject
-    protected ErrorHandler(ErrorHandlingCatalog errorHandlingCatalog)
+    public ErrorResponseFactory()
```
bannmann, line 34: We no longer need a constructor; move the assignments to the field declarations.

core/src/main/java/com/github/mizool/core/rest/errorhandling/ClientErrorMapper.java
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright 2018 incub8 Software Labs GmbH
```
bannmann, line 2: Extend the copyright period in all files touched to 2020.

core/src/main/java/com/github/mizool/core/rest/errorhandling/ErrorHandlingBehaviorCatalog.java
```diff
@@ -10,13 +26,13 @@
 import com.google.common.collect.ImmutableMap;
 
 @Singleton
-class ErrorHandlingCatalog
+class ErrorHandlingBehaviorCatalog
```
bannmann, line 29: Speaking of "spring cleanup": while you're at it, please also incorporate #147 in this PR.

core/src/main/java/com/github/mizool/core/rest/errorhandling/RestExceptionMapper.java
```diff
@@ -16,29 +16,27 @@
...
 {
-    private final ErrorHandler errorHandler;
+    private final ErrorResponseFactory errorResponseFactory;
 
-    @Inject
-    public RestExceptionMapper(ErrorHandler errorHandler)
+    public RestExceptionMapper()
     {
-        this.errorHandler = errorHandler;
+        this.errorResponseFactory = new ErrorResponseFactory();
```
bannmann, line 33: move to field declaration

technology/web/src/main/java/com/github/mizool/technology/web/AbstractErrorHandlingFilter.java
```diff
@@ -18,29 +18,33 @@
...
     @Override
     public void doFilter(HttpServletRequest request, HttpServletResponse response, FilterChain chain)
         throws IOException
     {
         try
         {
-            TransactionalResponseWrapper transactionalResponseWrapper = new TransactionalResponseWrapper(response);
-            chain.doFilter(request, transactionalResponseWrapper);
-            transactionalResponseWrapper.commit();
+            chain.doFilter(request, response);
```
JanDoerrenhaus, line 47: Note for posterity: The transactional response wrapper is some of the oldest code we wrote. It came from "the usual" handling of responses in times of JSPs. Later, when we adapted streaming our responses, nobody thought of the fact that this actually completely breaks streaming. All responses are first built into the heap, then streamed from there.
Yes, this means that responses can abort mid-stream now, but you cannot have streaming and transactionality at the same time.

PR #168, opened 2020-06-07
carriage return (\r) is now stripped from key files

Yombu:
in fact this did not even compile on my local system (Windows 10) before changing the regex.


schroenser:
fyi: @bannmann

PR #169, opened 2020-07-28
Add FutureStreamJoiner

bannmann:


core/src/main/java/com/github/mizool/core/concurrent/FutureStreamJoiner.java
```diff
@@ -0,0 +1,195 @@
...
+@UtilityClass
+public class FutureStreamJoiner
+{
+    /**
+     * Provides fluent syntax for joining {@link ListenableFuture ListenableFutures}. <br>
+     * <br>
+     * See {@link FutureStreamJoiner} for details.
+     */
+    @NoArgsConstructor(access = AccessLevel.PRIVATE)
+    public class Listenable
```
schroenser, line 71: Add `static`.

schroenser: ....or does it inherit static from the `@UtilityClass` annotation? Hmmm.... in that case, nevermind.

core/src/main/java/com/github/mizool/core/concurrent/FutureStreamJoiner.java
```diff
@@ -0,0 +1,195 @@
...
+            throw new IllegalArgumentException("Concurrency limit must be greater than 0");
+        }
+    }
+
+    private Runnable consumeStream(Stream<Void> stream)
+    {
+        return new StreamConsumingWorker(stream);
+    }
+
+    @Value
```
schroenser, line 163: See my comment on the `ListenableFutureCallback` in the `BufferedStreamAdapter` on why I think `@Value` should not be used here.

Replace both annotations with a `@RequiredArgsConstructor` and make the field `private final`.

core/src/main/java/com/github/mizool/core/concurrent/FutureStreamJoiner.java
```diff
@@ -0,0 +1,195 @@
...
+        @Override
+        @SneakyThrows
+        public void run()
+        {
+            try
+            {
+                stream.forEach(whatever -> {
+                });
+            }
+            catch (UncheckedExecutionException t)
```
schroenser, line 178: Reading this, I wonder if the `BufferedStreamAdapter` should even expose the unchecked execution exception.

#### Full stack example
During the store operation, a `Stream<CompletableFuture<FooRecord>>` comes into existence.
This is then parallelized using the `BufferedStreamAdapter`, resulting in a `Stream<FooRecord>`.
This is then mapped to a `Stream<Foo>` and propagated towards the REST layer, where it's again mapped to a `Stream<FooDto>` and then serialized to JSON.

This means, exceptions, that occur during consumption of the stream, happen during serialization. That part of the code is totally unaware about buffering, parallelization and everything that comes with it.

#### Theory
Since the `BufferedStreamAdapter` returns a `Stream<V>`, it also hides that there are futures and threads involved. Thus, it should also hide all exceptions that come from using futures and threads.

#### Idea
Let's centralize handling of execution exceptions inside the `handleFutureResult` method.
A "future result" can either be a value or an exception, as indicated by the signature of the `handleFutureResult` method.
If it's an exception, is definitely _has_ qualities of "something you only get when dealing with futures", either because the original exception from inside the future was wrapped or because there was an exception during execution of the future.
In both cases, after `handleFutureResult` is finished, there should be no trace of "futures" left. Even the `ValueHolder` does not need to know about futures or special exceptions that occur when dealing with futures, as the `ValueHolder` is already positioned _after_ the future was finished and only helps us to transport a value or a (non-future-related) exception towards the `BlockingSpliterator`.

Based on that, this `forEach` (and all other terminal operations on streams, for that matter) would not have to deal with an `UncheckedExecutionException`. Why would a stream even throw such a thing? Abstruse.

core/src/main/java/com/github/mizool/core/concurrent/BufferedStreamAdapter.java
```diff
@@ -47,21 +52,82 @@
...
+        {
+            verifyBufferSize(bufferSize);
+
+            return new BufferedStreamAdapter<ListenableFuture<V>, V>(futures,
+                bufferSize,
+                executorService,
+                this::addListener).adapt();
+        }
+
+        private <E> void addListener(ListenableFuture<E> future, Listener<E> listener)
```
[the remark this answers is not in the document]
bannmann: Actually, I wanted to avoid reusing either F and V (the type parameters of BufferedStreamAdapter) anywhere - if a nested class or its members use the same letter, it will shadow the outer one, which is confusing and/or misleading.

I'll make this consistent by changing the V to an E in line 68.

core/src/main/java/com/github/mizool/core/concurrent/BufferedStreamAdapter.java
```diff
@@ -83,56 +149,40 @@ public ValueHolder(Throwable throwable)
...
             {
                 throw new UncheckedExecutionException(throwable);
             }
             return value;
         }
     }
 
-    @NoArgsConstructor(access = AccessLevel.PRIVATE)
-    private class Callback implements FutureCallback<V>
+    @Value
```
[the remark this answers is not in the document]
bannmann: Good point.

Regarding the nitpick: neither, I did not think about finality at all. I'll make all nested classes final.

core/src/main/java/com/github/mizool/core/concurrent/BufferedStreamAdapter.java
```diff
@@ -83,56 +149,40 @@ public ValueHolder(Throwable throwable)
...
-        }
-
-        private BooleanSupplier capacityAvailable()
-        {
-            return () -> runningFutures.get() + results.size() < bufferSize;
-        }
+        return () -> runningFutures.get() + results.size() < bufferSize;
     }
 
     private class BlockingSpliterator implements Spliterator<V>
```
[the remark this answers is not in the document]
bannmann: This is intentional as BlockingSpliterator uses various members of the surrounding class (semaphore, results, streamDepleted, runningFutures).

Making it static would require keeping an explicit reference to the BufferedStreamAdapter instance it is connected to (passed to the constructor). Given that the spliterator is an integral part of the BufferedStreamAdapter mechanics, that doesn't feel like a good fit.

core/src/main/java/com/github/mizool/core/concurrent/BufferedStreamAdapter.java
```diff
@@ -194,29 +244,43 @@ public int characteristics()
...
 
-    private BufferedStreamAdapter(Stream<ListenableFuture<V>> futures, int bufferSize, ExecutorService executorService)
+    private BufferedStreamAdapter(
+        Stream<F> futures, int bufferSize, ExecutorService executorService, ListenerAdder<F, V> listenerAdder)
     {
         this.futures = futures;
         this.bufferSize = bufferSize;
         this.executorService = executorService;
+        this.listenerAdder = listenerAdder;
         this.results = new LinkedList<>();
```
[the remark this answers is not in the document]
bannmann: Good catch. I'll go with RequiredArgsConstructor to minimize boilerplate.

The only arguments for not using `@RequiredArgsConstructor` and writing the constructor I can think of are:

- to decouple its argument order from member declaration order (but that is a moot point for a `private` constructor), and
- to make a nice "heading" separating the members from the methods (a style we don't like)

But I suspect this was no conscious decision at the time, but just leftovers from iterative development (maybe the constructor used to do more interesting stuff).

core/src/main/java/com/github/mizool/core/concurrent/BufferedStreamAdapter.java
```diff
@@ -194,29 +244,43 @@ public int characteristics()
...
     }
 
+    @FunctionalInterface
+    private interface ListenerAdder<T, E> extends BiConsumer<T, Listener<E>>
```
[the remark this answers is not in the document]
bannmann: Nested interfaces are implicitly `static`, so our ordering rules already say they should be above the non-static members of BufferedStreamAdapter. I'll fix this.

Using Listener<E> for the ListenableFutureCallback member is great, will do.

Historical note: I remember that I invented these interfaces halfway through the implementation, so maybe this was the plan all along and I just forgot to go through with it.

core/src/main/java/com/github/mizool/core/concurrent/BufferedStreamAdapter.java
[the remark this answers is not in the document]
schroenser: Hrm... I was on the brink of asking you to make the inner classes non-static, so you can use the actual `V` from the outer class. However, I don't like non-static inner classes either. So yeah, consistent non-shadowing naming it is.

core/src/main/java/com/github/mizool/core/concurrent/FutureStreamJoiner.java
bannmann: Yes indeed. `@UtilityClass` marks all members static, which includes nested classes. The Lombok IntelliJ plugin complains when you include the keyword here.

core/src/main/java/com/github/mizool/core/concurrent/FutureStreamJoiner.java
bannmann: You certainly make a good point here.

To sum this up and put it another way: the stream returned by `BufferedStreamAdapter` should, for all intents and purposes, feel like a stream that produces values in a **synchronous** way, just as if some algorithm would compute them. And exceptions in such computations would only ever throw regular, non-futurish exceptions - the worst that could happen to a stream consumer is that it encounters a `RuntimeException` (or subclass) wrapping a checked exception.

That said, this wrapping is the officially documented behavior of `BufferedStreamAdapter`. Thus, I'm hesitant to change it in a backwards-incompatible way (we may be allowed to do so by semantic versioning because we're not 1.0 yet, but still). Also note that `FutureStreamJoiner`, despite being built on top of `BufferedStreamAdapter`, would not be affected as the joint future it produces is specified (and unit-tested) to behave exactly like any other future (wrapping exceptions in an `ExecutionException`).

I suggest we carefully analyze all call sites of this class in relevant corporate code bases. I expect that at least some of them do explicit exception checking and unwrapping and could break if the exceptions they want to handle are suddenly thrown directly instead of being wrapped.

Let's discuss this in person, and maybe do the aforementioned analysis.

Also, as a first step, I'll try extending the existing unit test for exception wrapping so we have a solid foundation. Right now, it only throws a runtime exception, not the full suite of throwable kinds that the `FutureStreamJoiner` tests verify. I also suspect inconsistent behavior depending on which element of the stream caused the exception (like the original problem with `ListenableFutureCollector` that led to us writing `FutureStreamJoiner`).

schroenser: You are right: I did not take into account that the wrapped exceptions are not only the current behavior but also the documented behavior. I agree that we should check usages first before breaking the API.
That being said: adapting the call sites to my proposed behavior should generally simplify them, so the changes are most likely pretty easy. But they have to be identified and performed regardless.

schroenser: Since your calendar is the more... complicated one... please schedule a one hour meeting between the two of us for the purpose of checking implications of the proposed change. When that's scheduled, this conversation can be resolved.

core/src/main/java/com/github/mizool/core/concurrent/FutureStreamJoiner.java
```diff
@@ -35,29 +34,46 @@
...
+ * type of throwable (checked exception, runtime exception, error).</li>
  * </ul>
- * <h3>Examples</h3>
+ * <h3>Exceptions thrown by the original stream</h3>
+ * If the original stream throws an exception instead of producing a future, the joint future will behave the same way
+ * as with exceptional completion of futures: the exception will be wrapped inside an {@link ExecutionException}, e.g.
+ * when invoking {@link Future#get() get()}.<br>
+ * <br>
+ * <h3>Usage examples</h3>
  * This class is intended to be used as follows:<br>
```
schroenser, line 58: Get rid of this line as it's superfluous to the heading.

core/src/test/java/com/github/mizool/core/concurrent/ThrowingStreamHarness.java
```diff
@@ -0,0 +1,278 @@
...
+        @SneakyThrows
+        @Override
+        public void run()
+        {
+            String me = "Runnable #" + number;
+
+            log.info("{} running.", me);
+            if (shouldFail())
+            {
+                log.info("{} throws Exception.", me);
```
schroenser, line 67: Perhaps change this to `log.info("{} throws {}.", me, throwableClass.getName());`?

core/src/test/java/com/github/mizool/core/concurrent/ThrowingStreamHarness.java
```diff
@@ -0,0 +1,278 @@
...
+
+        TaskStreamFactory result = new TaskStreamFactory()
+        {
+            @Override
+            public Stream<Task> createStream(Class<? extends Throwable> desiredThrowableClass)
+            {
+                return IntStream.rangeClosed(1, count)
+                    .mapToObj(number -> {
+                        Task task = new Task(number, number == failingRunnableNumber ? desiredThrowableClass : null);
+                        Task t2 = new Task(1, desiredThrowableClass);
```
schroenser, line 166: Remove unused `Task t2`.

schroenser (changes requested):

core/src/test/java/com/github/mizool/core/concurrent/ThrowingStreamHarness.java
bannmann: Sure, why not. I went with `getSimpleName()`, though.

schroenser (approved):

PR #170, opened 2020-08-28
Type-safe config property access wrapper

bannmann:


core/src/main/java/com/github/mizool/core/password/Pbkdf2WithHmacSha1Hasher.java
```diff
@@ -67,12 +68,11 @@ public String toString()
 
     public static final String ALGORITHM_NAME = "PBKDF2WithHmacSHA1";
 
-    @VisibleForTesting
-    protected static final String ITERATIONS_PROPERTY_NAME = Pbkdf2WithHmacSha1Hasher.class.getName() + ".iterations";
+    private static final PropertyNode CONFIG = Config.systemProperties()
```
schroenser, line 71: Inline this, as it's only used once. The old extracted property name had the purpose to reduce nested calls. The new design chains calls, which I don't see as a problem.

core/src/test/java/com/github/mizool/core/configuration/TestConfig.java
```diff
@@ -0,0 +1,336 @@
...
+        assertThat(root.child("something").stringValue().obtain()).isEqualTo("interesting");
+        assertThat(root.child("greeting").stringValue().obtain()).isEqualTo("Hello, World!");
+    }
+
+    @Test
+    public void testSystemProperties()
+    {
+        RootNode root = Config.systemProperties();
+        String lineSeparator = root.child("line.separator").stringValue().obtain();
+        assertThat(lineSeparator).isNotEmpty();
```
schroenser, line 91: Since `obtain` throws an exception, change that to `read` and base the assertion on the presence of the value.

Also note that the `read` code already filters empty strings. That might be something you want to test separately.

core/src/test/java/com/github/mizool/core/configuration/TestConfig.java
```diff
@@ -0,0 +1,336 @@
...
+
+    @Test
+    public void testConfigPicksUpPropertiesChangedLater()
+    {
+        Properties properties = new Properties();
+        PropertyNode node = Config.from(properties).child("breaking");
+
+        properties.setProperty("breaking", "news");
+
+        assertThat(node.stringValue().obtain()).isEqualTo("news");
```
schroenser, line 102: I'd like to change this to make it even more ballsy: extract a variable for `node.stringValue()` before setting the property, to ensure that even the `Value` class is lazy.

Also you could extend the test to first test that `value.read()` returns an `Optional.absent()`, then setting the property and asserting the `value.obtain()` again.

schroenser: Since there are already tests that ensure the function of `obtain`, I'd recommend going with `read` for the other tests to avoid being distracted by the exception thrown by `obtain`.

technology/aws/src/main/java/com/github/mizool/technology/aws/dynamodb/Configuration.java
```diff
@@ -1,19 +1,36 @@
...
 public class Configuration
 {
-    private static final String DEFAULT_REGION = Regions.EU_CENTRAL_1.name();
+    private static final PropertyNode CONFIG = Config.systemProperties().child("aws");
 
     public Regions getAwsRegion()
     {
-        String region = System.getProperty("aws.region", DEFAULT_REGION);
-        return Regions.valueOf(region);
+        return CONFIG.child("region").convertedValue(Regions::valueOf).read().orElse(Regions.EU_CENTRAL_1);
```
schroenser, line 29: Extract `CONFIG.child("region").convertedValue(Regions::valueOf)` to a `REGION` constant.

technology/aws/src/main/java/com/github/mizool/technology/aws/dynamodb/Configuration.java
```diff
@@ -1,19 +1,36 @@
...
     }
 
     public String getEndpoint()
     {
-        return System.getProperty("aws.endpoint");
+        return CONFIG.child("endpoint").stringValue().obtain();
```
schroenser, line 34: Extract `CONFIG.child("endpoint").stringValue()` to a `ENDPOINT` constant.

schroenser (changes requested):

core/src/test/java/com/github/mizool/core/configuration/TestConfig.java
bannmann: > Also note that the read code already filters empty strings. That might be something you want to test separately.

This was in fact already covered by the `TestValue.testValues()` case called "empty". However, your comment made me realize that it was not obvious enough to readers, so I refactored it into its own test method.

schroenser (approved):

PR #177, opened 2020-11-18
Add UrlRef class

bannmann:


core/src/main/java/com/github/mizool/core/UrlRef.java
```diff
@@ -0,0 +1,185 @@
...
+
+    /**
+     * @throws NullPointerException if {@code spec} is null
+     * @throws IllegalArgumentException if the resulting {@code UrlRef} would cause exceptions when converted to {@link URI} or {@link URL}
+     */
+    public static UrlRef of(@NonNull String spec)
+    {
+        try
+        {
+            return new UrlRef(new URL(spec).toURI());
```
schroenser, line 153: Why not simply call `of(new URL(spec))` or `of(new URI(spec))`?

core/src/main/java/com/github/mizool/core/UrlRef.java
```diff
@@ -0,0 +1,185 @@
...
+ *     </li>
+ *     <li>
+ *         Offers a convenient way to build a child URL regardless if the base URL is a {@link URL}, a {@link URI} or a
+ *         {@code UrlRef}.
+ *     </li>
+ * </ul>
+ */
+@EqualsAndHashCode
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public final class UrlRef
```
schroenser, line 57: Why not simply store a `private final String spec;`, add a constructor that uses two `verify` calls to ensure that it's both a valid `URI` _and_ `URL` and have the `to`ers create instances on the fly?

This class would be _way_ shorter, much simpler and `equals` and `hash` code work brilliantly with a `String` field.

schroenser, line 57: Also, I don't like the
* `of(URI, String)`,
* `of(URL, String)` and
* `of(UrlRef, String)`
factory methods.

I'd rather see a `public UrlRef resolve(String spec)` as an instance method.

`UrlRef.of(myUrlRef, "/newpath")` would become `myUrlRef.resolve("/newPath")`.

schroenser, line 57: Specialized constructors that take `URL` and `URI` would also be fine. However, I would omit the constructors that take an additional `String spec`. Instead of `new UrlRef(myURL, "/newPath")` the code would be `new UrlRef(myURL).resolve("/newPath")`.

schroenser, line 57: I don't know how I feel about `new UrlRef(URL)`, `new UrlRef(URI)` and `new UrlRef(String)` vs. `UrlRef.of(URL)`, `UrlRef.of(URI)` and `UrlRef.of(String)`.
I don't see the benefit of the factory methods. At least in this case, the constructor calls are quite similar in terms of code length/readability.

schroenser, line 57: Combining the comments above, my proposal for the public API of this class would be
* `public UrlRef(String spec)`
* `public UrlRef(URL url)`
* `public UrlRef(URI uri)`
* `public UrlRef resolve(String spec)`
* `public URL toUrl()`
* `public URI toURI()`
* `public String toString()`

schroenser (changes requested):

core/src/main/java/com/github/mizool/core/UrlRef.java
bannmann: Good points. I guess I just stuck with overloaded factory methods I started with, never noticing that a single `resolve(String spec)` would abolish the need for having them in the first place.

schroenser (approved):

PR #180, opened 2020-12-20
Fix error handling for non-JAX-RS code

bannmann:
- RuntimeExceptions thrown by non-JAX-RS code (e.g. Shiro filters/realms, injection problems) were not caught, causing servlet container error pages.
- Bug was present since Mizool 0.38, caused by 2bd09aa11480a8c2df5ed6d99ab2fffd59e8ca7d.
- Note: JAX-RS code was not affected because it is covered by RestExceptionMapper.

I also removed the dead TransactionalResponseWrapper class and made AbstractErrorHandlingFilter friendlier for subclasses.


bannmann:
FYI @JanDoerrenhaus @schroenser

PR #183, opened 2021-03-19
Feature/update password hashing algorithm

voons:


core/src/main/java/com/github/mizool/core/password/PasswordHasherRegistry.java
```diff
@@ -31,8 +37,27 @@
...
+        Optional<PasswordHasher> defaultHasher = defaultPasswordHashers.stream().findFirst();
+        if (!defaultHasher.isPresent())
+        {
+            throw new CodeInconsistencyException(
+                "No hasher instances for PasswordHasher found, one should be registered with META-INF services");
+        }
+        if (defaultPasswordHashers.size() > 1)
+        {
+            throw new CodeInconsistencyException(
+                "Found several hasher instances for PasswordHasher, only one should be registered with META-INF services");
```
bannmann, line 59: This won't work. To my knowledge, applications cannot reliably un-register a service that is registered by a jar file on their classpath.

Let's change this as follows:

1. _All_ hasher implementations register via `@MetaInfServices`
2. Among all those hashers, `getDefaultHasher()` finds the one whose name matches the one given by a system property `com.github.mizool.core.password.PasswordHasherRegistry.defaultAlgorithm`. If it is unset, the registry defaults to `Pbkdf2WithHmacSha512Hasher.ALGORITHM_NAME`.
3. PasswordHashTool always iterates through all hashers and displays each digest (allowing the developer to pick the algorithm in use by their application)

core/src/main/java/com/github/mizool/core/password/PasswordHasherRegistry.java
```diff
@@ -31,8 +37,27 @@
...
-        return Iterables.getOnlyElement(
-            Iterables.filter(
-                passwordHashers, passwordHasher -> passwordHasher.isResponsibleFor(algorithm)));
+        return Iterables.getOnlyElement(Iterables.filter(passwordHashers,
+            passwordHasher -> passwordHasher.isResponsibleFor(algorithm)));
+    }
+
+    public static PasswordHasher getDefaultHasher()
+    {
+        Iterable<PasswordHasher> hasherInstances = MetaInfServices.instances(PasswordHasher.class);
```
bannmann, line 46: Don't interact with `MetaInfServices`, use the injected `Instance<PasswordHasher> passwordHashers` to stay compatible with CDI-based applications. For non-CDI environments like `PasswordHashTool`, let's use an MetaInfServices-to-Instance adapter. (I don't know offhand if we already have one in this code base, contact me if not and I'll look it up elsewhere.)

bannmann (changes requested):

tool/password/src/main/java/com/github/mizool/tool/password/PasswordHashTool.java
```diff
@@ -39,14 +43,18 @@ public static void main(String[] args)
             }
             else
             {
-                PasswordHashTool passwordHashTool = new PasswordHashTool(PasswordHasherRegistry.getDefaultHasher(),
-                    parameters.getPassword());
+                ServiceLoaderInstanceImpl<PasswordHasher>
+                    instances
+                    = new ServiceLoaderInstanceImpl<>(PasswordHasher.class);
```
bannmann, line 48: Okay, I now realize that in proposing `ServiceLoaderInstanceImpl`, I blindly assumed the CLI tool would use the `PasswordHasherRegistry` and would therefore need to supply an `Instance`. While that would work, it would mean adding a `getAllHashers()` method to the registry that arguably no other call site would ever need.

The code right now is a mixture of two approaches, and to be honest, I'm undecided between them. Please choose one of the following approaches:

1. Encapsulate instantiation of the `Instance<PasswordHasher>`: add Guice, add a Guice module class that knows how to provide an `Instance<PasswordHasher>`, then have `PasswordHashTool` ask Guice for a ready-made instance of `PasswordHasherRegistry`, then invoke a newly added method `getAllHashers()` on it.
2. Simplify things because we're not using the registry anyway: remove the `ServiceLoaderInstanceImpl`, then have `PasswordHashTool` call `MetaInfServices.instances()`. 
    - Note that this is different from the original commit where the registry interacted with `MetaInfServices`. Now, the registry is usable for both CDI and non-CDI environments and - in the latter case - leaves the application in charge of supplying an implementation of `Instance<PasswordHasher>`. `PasswordHashTool`, on the other hand, is not interested in the registry, so it doesn't need an `Instance<PasswordHasher` and can be safely hard-coded to use `MetaInfServices`.

bannmann (changes requested):

core/src/main/java/com/github/mizool/core/password/Pbkdf2WithHmacSha512Hasher.java
```diff
@@ -0,0 +1,49 @@
...
+public class Pbkdf2WithHmacSha512Hasher extends Pbkdf2WithHmacShaHasher
+{
+    public static final String ALGORITHM_NAME = "PBKDF2WithHmacSHA512";
+
+    private static final int ITERATIONS_FOR_NEW_PASSWORDS = Config.systemProperties()
+        .child(Pbkdf2WithHmacSha512Hasher.class.getName())
+        .child("iterations")
+        .intValue()
+        .read()
+        .orElse(121212);
```
bannmann, line 36: Change to `121_212` to make the number more readable (see [SonarCloud issue](https://sonarcloud.io/project/issues?id=com.github.mizool%3Amizool&issues=AXhKZ--pBOoP0w3KeMvj&open=AXhKZ--pBOoP0w3KeMvj&pullRequest=183)).

As this is a value that was found through experiments, why not round it (up/down) to a value that looks less magical/funny, e.g. `122_000` or the like?

core/src/main/java/com/github/mizool/core/password/Pbkdf2WithHmacSha1Hasher.java
```diff
@@ -16,57 +16,17 @@
...
+import org.kohsuke.MetaInfServices;
+
 import com.github.mizool.core.configuration.Config;
-import com.github.mizool.core.exception.CodeInconsistencyException;
-import com.google.common.base.Joiner;
-import com.google.common.base.Splitter;
 
 @Slf4j
-public class Pbkdf2WithHmacSha1Hasher implements PasswordHasher
+@Deprecated
```
bannmann, line 26: The Javadoc `@deprecated` tag is missing (see [SonarCloud issue](https://sonarcloud.io/project/issues?id=com.github.mizool%3Amizool&issues=AXhKZ-72BOoP0w3KeMvg&open=AXhKZ-72BOoP0w3KeMvg&pullRequest=183)).

This gives us an opportunity to explain _why_ this is deprecated. How about something like this? `For newly saved passwords, use {@link Pbkdf2WithHmacSha512Hasher} instead. This hasher is kept to allow continued verification of old passwords.`

core/src/main/java/com/github/mizool/core/password/Pbkdf2WithHmacShaHasher.java
```diff
@@ -0,0 +1,137 @@
...
+
+import lombok.Data;
+import lombok.extern.slf4j.Slf4j;
+
+import com.github.mizool.core.exception.CodeInconsistencyException;
+import com.google.common.base.Joiner;
+import com.google.common.base.Splitter;
+
+@Slf4j
+public abstract class Pbkdf2WithHmacShaHasher implements PasswordHasher
```
bannmann, line 38: Rename to `AbstractPbkdf2WithHmacShaHasher`.

tool/password/src/main/java/com/github/mizool/tool/password/PasswordHashTool.java
```diff
@@ -39,12 +43,16 @@ public static void main(String[] args)
...
-                if (digest != null)
-                {
-                    System.out.println("Digest: " + digest);
-                }
+                Iterable<PasswordHasher> instances = MetaInfServices.instances(PasswordHasher.class);
+                instances.forEach(hasher -> {
+                    PasswordHashTool passwordHashTool = new PasswordHashTool(hasher, parameters.getPassword());
+
+                    String digest = passwordHashTool.getDigest();
+                    if (digest != null)
```
bannmann, line 51: I have no idea why we added this null check here. If a hasher _did_ have trouble with a password, it should throw an exception. Having it return `null` and silently skip it is not the way to go. Remove the check and inline the `digest` variable.

tool/password/src/main/java/com/github/mizool/tool/password/PasswordHashTool.java
```diff
@@ -39,12 +43,16 @@ public static void main(String[] args)
...
             else
             {
-                PasswordHashTool passwordHashTool = new PasswordHashTool(parameters.getPassword());
-                String digest = passwordHashTool.getDigest();
-                if (digest != null)
-                {
-                    System.out.println("Digest: " + digest);
-                }
+                Iterable<PasswordHasher> instances = MetaInfServices.instances(PasswordHasher.class);
+                instances.forEach(hasher -> {
```
bannmann, line 47: While we're cleaning up things: there's no reason to instantiate the `PasswordHashTool` - its instance methods only consist of one-liners that could be equally placed at the call site. Therefore, drop fields, constructor and instance methods.

Still, having some separation between high-level code in `main` and other stuff is useful: let's replace this lambda with a reference to a new static method `void printDigest(PasswordHasher hasher)`.

bannmann (changes requested):

tool/password/src/main/java/com/github/mizool/tool/password/PasswordHashTool.java
```diff
@@ -39,12 +43,16 @@ public static void main(String[] args)
...
+
+                    String digest = passwordHashTool.getDigest();
+                    if (digest != null)
+                    {
+                        log.info("Algorithm: {}  Digest: {}", passwordHashTool.getAlgorithmName(), digest);
+                    }
+                });
             }
         }
         catch (ParameterException ignored)
```
bannmann, line 58: `log.debug()` the exception, rename variable to `e`

bannmann (approved):

PR #184, opened 2021-03-25
Add Synchronizer to replace low-level semaphore helper methods

bannmann:


core/src/main/java/com/github/mizool/core/concurrent/Threads.java
```diff
@@ -1,6 +1,6 @@
 /*
- * Copyright 2018 incub8 Software Labs GmbH
- * Copyright 2018 protel Hotelsoftware GmbH
+ * Copyright 2018-2020 incub8 Software Labs GmbH
```
bannmann, line 2: Note to self: bump to 2021

core/src/main/java/com/github/mizool/core/concurrent/SynchronizerApi.java
```diff
@@ -0,0 +1,163 @@
+package com.github.mizool.core.concurrent;
```
bannmann, line 1: Note to self: add license boilerplate

core/src/main/java/com/github/mizool/core/concurrent/ListenableFutureCollector.java
```diff
@@ -60,7 +59,7 @@
  * stream.map(ResultVoidingFuture::new).collect(ListenableFutureCollector.concurrent(maximumConcurrentFutures)).get();}</pre>
  *
  * @deprecated This class does not propagate exceptions consistently and will be removed. Use {@link FutureStreamJoiner}
- *     instead.
+ * instead.
  */
 @Deprecated
```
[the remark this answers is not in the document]
bannmann: Mizool targets Java 8, but `forRemoval` was added in Java 9.

core/src/main/java/com/github/mizool/core/concurrent/SynchronizerApi.java
bannmann: @lesmou and I have already scheduled a kickoff for May 21 on this (yes, that far into the future, thanks to various vacations).

schroenser: I don't care about the far-ness into the future-ness. But I'm done with "yeah, let's do this some time soon" and then not doing it.

A scheduled meeting is already an improvement as it puts meaning behind the idea, regardless of when it will happen.

schroenser (approved):

schroenser:
fyi @JanDoerrenhaus

Note that I can't mention Georg, as he hasn't accepted his team invite yet.

PR #187, opened 2021-04-20
Added stream reduction operation for streams with equal values.

schroenser:


core/src/main/java/com/github/mizool/core/Streams.java
```diff
@@ -33,4 +37,22 @@
...
+     * <br>
+     * A {@link ReductionException} will be thrown during reduction if one or more of the values in the {@link Stream}
+     * are not equal according to {@link Objects#equals(Object, Object)}.
+     */
+    public <T> BinaryOperator<T> expectingEqualValues()
+    {
+        return (t, u) -> {
+            if (!Objects.equals(t, u))
+            {
+                throw new ReductionException("\"" + u.toString() + "\" is not equal to \"" + t.toString() + "\"");
```
JanDoerrenhaus, line 53: While the `Objects.equals` call is null safe, the `.toString()` calls are not. Use `Objects.toString` instead.

JanDoerrenhaus (changes requested):

core/src/main/java/com/github/mizool/core/Streams.java
schroenser: Good catch!

Interestingly enough, IntelliJ's inspection then says that `Objects.toString(u)` can be replaced with `u` because both will result in the same behaviour (concatenating `"null"` if `u == null` and `u.toString()` if `u != null`.

I'll replace both without any method calls.

JanDoerrenhaus: > Interestingly enough, IntelliJ's inspection then says that `Objects.toString(u)` can be replaced with `u` because both will result in the same behaviour (concatenating `"null"` if `u == null` and `u.toString()` if `u != null`.

That would be true if you did `"\"" + u + "\" is not equal to \"" + t + "\""`. The method calls are definitely null unsafe.

schroenser: Yeah, that's what I meant. Brain mushy, English no good. :-P

schroenser:
fyi: @bannmann

PR #191, opened 2021-05-17
Made use of the proper ConflictingEntityException and ObjectNotFoundE…

JanDoerrenhaus:
…xception on those use cases, and introduced try-with-resources for the used DslContext.

### DIFF
--- a/technology/sql/src/main/java/com/github/mizool/technology/sql/jooq/Records.java
+++ b/technology/sql/src/main/java/com/github/mizool/technology/sql/jooq/Records.java
@@ -18,12 +18,15 @@ package com.github.mizool.technology.sql.jooq;
 
 import lombok.experimental.UtilityClass;
 
+import org.jooq.CloseableDSLContext;
 import org.jooq.Condition;
 import org.jooq.Table;
 import org.jooq.TableRecord;
 import org.jooq.UpdatableRecord;
 import org.jooq.exception.DataAccessException;
 
+import com.github.mizool.core.exception.ConflictingEntityException;
+import com.github.mizool.core.exception.ObjectNotFoundException;
 import com.github.mizool.core.exception.StoreLayerException;
 
 @UtilityClass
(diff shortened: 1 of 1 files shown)


technology/sql/src/main/java/com/github/mizool/technology/sql/jooq/Records.java
```diff
@@ -89,12 +96,16 @@
...
-    public <R extends UpdatableRecord<R>> int delete(
+    public <R extends UpdatableRecord<R>> void delete(
         R record, Table<R> table, MizoolConnectionProvider mizoolConnectionProvider)
     {
-        try
+        try (CloseableDSLContext dslContext = DslContexts.obtain(mizoolConnectionProvider))
         {
-            return DslContexts.obtain(mizoolConnectionProvider).executeDelete(record);
+            int deletedRows = dslContext.executeDelete(record);
+            if (deletedRows != 1)
```
lesmou, line 105: This is changing the behaviour in situations where we want to delete rows but none are present.
I am still not comfortable with that approach to handle these situations as erroneous since during a regular cleanup there might be a delete operation on already empty tables. I recall having that discussion already once but do not remember the outcome.
What are your thoughts? @JanDoerrenhaus @bannmann

lesmou (changes requested):
Please add to discussion.

technology/sql/src/main/java/com/github/mizool/technology/sql/jooq/Records.java
JanDoerrenhaus: So far, we have been following the pattern, that if you DELETE an item that does not exist, a 404 is returned, not a 200. This class is meant to be used for the standard CRUD operations of REST stacks, not generically for each and every SQL delete that may ever happen. So if I see a "regular cleanup" using this class, I would question the sanity of the author.

bannmann: What Jan said.

Sure, there are cases where we don't care if rows were already deleted (or never existed in the first place). For those cases, we simply create a separate store with a behavior-revealing name (maybe `FooCleanupStore` or `FooEnsureNoneExistStore` for a `Foo` table).

lesmou (approved):

bannmann (approved):

bannmann:
FYI @gregorxiv

PR #192, opened 2021-06-24
updated Identifiable interface with default implementations (MIZOOL-161)

voons:


core/src/main/java/com/github/mizool/core/Identifiable.java
```diff
@@ -18,5 +18,13 @@
 
 public interface Identifiable<T extends Identifiable>
 {
-    Identifier<T> getIdentifier();
+    default Identifier<T> getIdentifier()
```
bannmann, line 21: Add an `@Deprecated` annotation plus an `@deprecated` Javadoc tag that says to use getId() instead.

bannmann (changes requested):

bannmann (approved):

PR #193, opened 2021-07-19
Mizool Relicensing

bannmann:


bannmann:
Note: we found that relicensing was simpler than we thought as the only code copied from elsewhere was removed in 2017 with commit 184f9482d3be59294996ce6dee2653aabcb30917.

bannmann:
@schroenser: @lesmou suggested that for this particular PR, we add Yombu to the list of people who are FYI'd.

schroenser (approved):

schroenser:
Note that I've chosen to ignore the 14 Code Smells, as they are not introduced by this change.

schroenser:
FYI: @gregorxiv @JanDoerrenhaus @Yombu

PR #194, opened 2021-08-12
Add ParameterizedThrowable, use it in GenericErrorMapper to add param…

matejjj18:
Add ParameterizedThrowable, use it in GenericErrorMapper to add parameters from getExceptionParameters to ErrorDto

### DIFF
--- a/core/src/main/java/com/github/mizool/core/rest/errorhandling/GenericErrorMapper.java
+++ b/core/src/main/java/com/github/mizool/core/rest/errorhandling/GenericErrorMapper.java
@@ -89,9 +89,9 @@ public class GenericErrorMapper
         Map<String, Object> parameters = Maps.newHashMap();
         parameters.put("Exception", throwable.getMessage());
         parameters.put("RootCause", Throwables.getRootCause(throwable).getMessage());
-        if (throwable instanceof ParameterizedThrowable)
+        if (throwable instanceof ParameterizedException)
         {
-            Map<String, Object> exceptionParameters = ((ParameterizedThrowable) throwable).getExceptionParameters();
+            Map<String, Object> exceptionParameters = ((ParameterizedException) throwable).getExceptionParameters();
             if (exceptionParameters != null)
             {
                 parameters.putAll(exceptionParameters);
--- a/core/src/main/java/com/github/mizool/core/rest/errorhandling/GenericErrorMapper.java
+++ b/core/src/main/java/com/github/mizool/core/rest/errorhandling/GenericErrorMapper.java
@@ -34,7 +34,7 @@ public class GenericErrorMapper
     public ErrorResponse handleUndefinedError(Throwable throwable)
     {
         log.error("Unhandled error", throwable);
-        Map<String, String> parameters = createExceptionParameters(throwable);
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/github/mizool/core/rest/errorhandling/GenericErrorMapper.java
```diff
@@ -84,11 +84,19 @@ private Throwable determineRootCause(Throwable t)
...
+    private Map<String, Object> createExceptionParameters(Throwable throwable)
     {
-        Map<String, String> parameters = Maps.newHashMap();
+        Map<String, Object> parameters = Maps.newHashMap();
         parameters.put("Exception", throwable.getMessage());
         parameters.put("RootCause", Throwables.getRootCause(throwable).getMessage());
+        if (throwable instanceof ParameterizedThrowable)
+        {
+            Map<String, Object> exceptionParameters = ((ParameterizedThrowable) throwable).getExceptionParameters();
+            if (exceptionParameters != null)
```
schroenser, line 95: I was thinking about whether we should allow the parameters to be `null`. Since the developer made the effort to implement `ParameterizedThrowable`, it could be safe to assume that `null` parameters are some kind of `CodeInconsistencyException`.

However, I really don't want to add exception cases to the exception handling, so perhaps this is the solution with the least amount of pain down the road.

What do you think @bannmann ?

core/src/main/java/com/github/mizool/core/rest/errorhandling/ParameterizedThrowable.java
```diff
@@ -0,0 +1,8 @@
+package com.github.mizool.core.rest.errorhandling;
+
+import java.util.Map;
+
+public interface ParameterizedThrowable
```
schroenser, line 5: I get where the name is derived from as the `GenericErrorMapper` handles `Throwable` instead of `Exception`.
I remember that it was necessary for the error mapper to be able to handle `Throwable`, so as far as I can remember, we can't change that.

However I think we shouldn't encourage developers to extend `Throwable` ever. Although this is only an interface, the name is suggestive.

Rename this to `ParameterizedException`.

Yes, technically it _could_ be applied to a `Throwable` as well and would still work. I really hope that never happens, regardless of having a good reason or not.

schroenser (changes requested):

core/src/main/java/com/github/mizool/core/rest/errorhandling/GenericErrorMapper.java
bannmann: I agree. Further rationale: the main objective is to transport the original error. So even if we suspect that it is likely lacking some details, we should not *replace* that error with a `CodeInconsistencyException`.

Also, I wouldn't bother *adding* information about that case because the client / end user is likely not able to fix it, and it's not their business anyway.

As an alternative, I briefly considered adding a log warning here, but decided against it to avoid log spam in production systems. If the lack of details is relevant, (frontend) developers will surely notice the error anyway.

schroenser (approved):

bannmann (approved):

bannmann:
FYI @gregorxiv @JanDoerrenhaus

PR #195, opened 2021-08-16
Make UnprocessableEntityException deprecated with appropriate message in javadoc

matejjj18:
… in javadoc

### DIFF
--- a/core/src/main/java/com/github/mizool/core/exception/AbstractUnprocessableEntityException.java
+++ b/core/src/main/java/com/github/mizool/core/exception/AbstractUnprocessableEntityException.java
@@ -1,0 +1,29 @@
+package com.github.mizool.core.exception;
+
+/**
+ * Thrown when the entity is syntactically correct (e.g. all non-null properties are set), but could not be processed
+ * for semantic reasons (e.g. invalid references to other database entities).
+ * Extend this class to create an exception class specific to the respective problem.
+ */
+public abstract class AbstractUnprocessableEntityException extends RuntimeException
+{
+    public AbstractUnprocessableEntityException()
+    {
+        super();
+    }
+
+    public AbstractUnprocessableEntityException(String message, Throwable cause)
+    {
+        super(message, cause);
+    }
+
+    public AbstractUnprocessableEntityException(String message)
+    {
+        super(message);
+    }
+
+    public AbstractUnprocessableEntityException(Throwable cause)
+    {
+        super(cause);
+    }
+}
--- a/core/src/main/java/com/github/mizool/core/exception/AbstractUnprocessableEntityException.java
+++ b/core/src/main/java/com/github/mizool/core/exception/AbstractUnprocessableEntityException.java
@@ -7,22 +7,22 @@ package com.github.mizool.core.exception;
  */
 public abstract class AbstractUnprocessableEntityException extends RuntimeException
 {
-    public AbstractUnprocessableEntityException()
+    protected AbstractUnprocessableEntityException()
     {
         super();
     }
 
-    public AbstractUnprocessableEntityException(String message, Throwable cause)
+    protected AbstractUnprocessableEntityException(String message, Throwable cause)
     {
         super(message, cause);
     }
 
-    public AbstractUnprocessableEntityException(String message)
+    protected AbstractUnprocessableEntityException(String message)
     {
         super(message);
     }
 
-    public AbstractUnprocessableEntityException(Throwable cause)
+    protected AbstractUnprocessableEntityException(Throwable cause)
     {
         super(cause);
     }
--- a/core/src/main/java/com/github/mizool/core/exception/UnprocessableEntityException.java
+++ b/core/src/main/java/com/github/mizool/core/exception/UnprocessableEntityException.java
@@ -1,8 +1,11 @@
 package com.github.mizool.core.exception;
 
 /**
- * Thrown when the entity is syntactically correct (e.g. all non-null properties are set), but could not be processed
- * for semantic reasons (e.g. invalid references to other database entities).
+ * @deprecated It is designed to be thrown when the entity is syntactically correct (e.g. all non-null properties are set),
+ * but could not be processed for semantic reasons (e.g. invalid references to other database entities).
+ * But the UnprocessableEntityException doesn't really say why the entity is unprocessable and
+ * is generally a bad fit as no client ever really has a chance of handling such a response. Try to avoid using this
+ * exception and use more specific one.
  */
 public class UnprocessableEntityException extends RuntimeException
 {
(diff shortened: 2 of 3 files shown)


core/src/main/java/com/github/mizool/core/exception/UnprocessableEntityException.java
```diff
@@ -1,9 +1,13 @@
 package com.github.mizool.core.exception;
 
 /**
- * Thrown when the entity is syntactically correct (e.g. all non-null properties are set), but could not be processed
- * for semantic reasons (e.g. invalid references to other database entities).
+ * @deprecated It is designed to be thrown when the entity is syntactically correct (e.g. all non-null properties are set),
+ * but could not be processed for semantic reasons (e.g. invalid references to other database entities).
+ * But the UnprocessableEntityException doesn't really say why the entity is unprocessable and
+ * is generally a bad fit as no client ever really has a chance of handling such a response. Try to avoid using this
+ * exception and use more specific one.
```
schroenser, line 8: _(...) exception and use_ **a** _more specific one._

core/src/main/java/com/github/mizool/core/exception/UnprocessableEntityException.java
```diff
@@ -1,9 +1,13 @@
...
  */
+@Deprecated
```
schroenser, line 10: Although it's the default value, add `(forRemoval = false)` to indicate that we consciously decided to keep the class but simply discourage its use.

schroenser (changes requested):

core/src/main/java/com/github/mizool/core/exception/UnprocessableEntityException.java
matejjj18: `(forRemoval = false)` is available only from Java 9

schroenser: It's like being in the dark ages....

schroenser (approved):

core/src/main/java/com/github/mizool/core/exception/UnprocessableEntityException.java
bannmann: When discussing this, my idea was that the more specific exceptions that applications define should extend this class. This way, one would not have to define exception behaviors each time one creates such an exception.

Deprecating the class would create compiler warnings in all those concrete exceptions due to `extends UnprocessableEntityException`. We could then of course add a `@SuppressWarnings("deprecation")`, but that feels quite weird because we don't want to discourage _extending the class_, but discourage _creating instances of it_.

My next thought was to deprecate the constructors instead of the class. However, that still doesn't quite work because then the `super(...)` calls in the constructors of subclasses would trigger compiler warnings.

So it's not easy. Still, I very much would like to have one base class to make it really convenient to create specific exceptions.

So here's my next idea: How about we keep the deprecation of this class, but introduce a new class `AbstractUnprocessableEntityException`?

- The existing exception behavior would be changed to use  `AbstractUnprocessableEntityException` instead.
- Both the deprecated not-specific-enough `UnprocessableEntityException` and any properly-specific application-provided `FooException` extend  `AbstractUnprocessableEntityException`.

core/src/main/java/com/github/mizool/core/exception/UnprocessableEntityException.java
```diff
@@ -1,9 +1,13 @@
 package com.github.mizool.core.exception;
 
 /**
- * Thrown when the entity is syntactically correct (e.g. all non-null properties are set), but could not be processed
- * for semantic reasons (e.g. invalid references to other database entities).
+ * @deprecated It is designed to be thrown when the entity is syntactically correct (e.g. all non-null properties are set),
```
bannmann, line 4: Not having a class description is bad style, and "sabotages" things like package overview tables in Javadoc and IDE tooltips.

Your deprecation comment is quite good, though. I just rephrased it a bit and incorporated the abstract class mentioned in my comment below.

> `@deprecated Throwing UnprocessableEntityException doesn't say <i>why</i> the entity is unprocessable and is generally a bad fit as no client ever really has a chance of handling such a response. Instead, consider creating an exception class specific to the respective problem that extends {@link AbstractUnprocessableEntityException}.`

bannmann (changes requested):

core/src/main/java/com/github/mizool/core/exception/UnprocessableEntityException.java
matejjj18: I think this is good way to go. That way we
- will avoid extending UnprocessableEntityException which would cause all compiler warnings
- support extending and discourage creating instances of UnprocessableEntityException
- will be able to completely get rid of UnprocessableEntityException in the future

schroenser: I like.

core/src/main/java/com/github/mizool/core/exception/AbstractUnprocessableEntityException.java
```diff
@@ -0,0 +1,29 @@
+package com.github.mizool.core.exception;
+
+/**
+ * Thrown when the entity is syntactically correct (e.g. all non-null properties are set), but could not be processed
+ * for semantic reasons (e.g. invalid references to other database entities).
+ * Extend this class to create an exception class specific to the respective problem.
+ */
+public abstract class AbstractUnprocessableEntityException extends RuntimeException
+{
+    public AbstractUnprocessableEntityException()
```
bannmann, line 10: Please fix [the SonarCloud issue](https://sonarcloud.io/project/issues?id=com.github.mizool%3Amizool&issues=AXtyA9xirRq0z5t8KDsJ&open=AXtyA9xirRq0z5t8KDsJ&pullRequest=195) for this and the other constructors (protected instead of public).

core/src/main/java/com/github/mizool/core/rest/errorhandling/behavior/UnprocessableEntityExceptionBehavior.java
```diff
@@ -13,7 +13,7 @@
     @Override
     public Class<? extends Throwable> getThrowableClass()
     {
-        return UnprocessableEntityException.class;
+        return AbstractUnprocessableEntityException.class;
```
bannmann, line 16: If you didn't do so already, please double-check that this indeed still works for the `UnprocessableEntityException` subclass. (I'm not sure whether we already had this use case.)

bannmann (changes requested):

core/src/main/java/com/github/mizool/core/rest/errorhandling/behavior/UnprocessableEntityExceptionBehavior.java
matejjj18: Test com.github.mizool.core.rest.errorhandling.TestErrorHandlingBehaviorCatalog#whitelistedBySuperclass proves it works

bannmann (approved):

bannmann:
FYI @gregorxiv @JanDoerrenhaus

PR #196, opened 2021-08-20
Just injecting whichever executor service the application might provi…

JanDoerrenhaus:
…de to do caching turns out to be dangerous, as it can lead to deadlocks when asynchronous processes access the cache, so now the TimeoutingExecutor gets its own thread pool.


technology/cache/src/main/java/com/github/mizool/technology/jcache/timeouting/TimeoutingExecutor.java
```diff
@@ -41,51 +33,16 @@
...
-            Thread.currentThread()
-                .interrupt();
-            throw new UncheckedInterruptedException(e);
-        }
-        catch (ExecutionException e)
-        {
-            throw new UncheckedExecutionException(e);
-        }
+        return Futures.get(executorService.submit(callable),
+            Duration.of(TimeoutingExecutor.CACHE_TIMEOUT, ChronoUnit.MILLIS));
```
bannmann, line 41: Let's catch `UncheckedTimeoutException` and wrap it in a `CacheTimeoutException`. This way, we keep the existing behavior that `SafeCacheLogHelper` relies upon.

Alternatively, we should remove `CacheTimeoutException` and change `SafeCacheLogHelper` to catch `UncheckedTimeoutException` instead.

I tend towards the second solution.

technology/cache/src/main/java/com/github/mizool/technology/jcache/timeouting/TimeoutingExecutor.java
```diff
@@ -41,51 +33,16 @@
...
-        {
-            Thread.currentThread()
-                .interrupt();
-            throw new UncheckedInterruptedException(e);
-        }
-        catch (ExecutionException e)
-        {
-            throw new UncheckedExecutionException(e);
-        }
+        Futures.get(executorService.submit(runnable), Duration.of(TimeoutingExecutor.CACHE_TIMEOUT, ChronoUnit.MILLIS));
```
bannmann, line 46: Same as above.

bannmann (changes requested):

bannmann (approved):

bannmann:
FYI @gregorxiv

PR #198, opened 2021-08-23
Feature/prsd 585 add a traceid to mizools errorresponse

schroenser:


schroenser:
There was some automatic code formatting going on. I've made the effort to split the formatting into a separate commit, so the real change might be easier to spot in the second commit.

core/src/main/java/com/github/mizool/core/rest/errorhandling/ErrorResponseFactory.java
```diff
@@ -10,10 +10,23 @@
...
-    private final ConstraintViolationMapper constraintViolationMapper = new ConstraintViolationMapper();
-    private final RuleViolationMapper ruleViolationMapper = new RuleViolationMapper();
-    private final ClientErrorMapper clientErrorMapper = new ClientErrorMapper();
-    private final GenericErrorMapper genericErrorMapper = new GenericErrorMapper();
+    private final ConstraintViolationMapper constraintViolationMapper;
+    private final RuleViolationMapper ruleViolationMapper;
+    private final ClientErrorMapper clientErrorMapper;
+    private final GenericErrorMapper genericErrorMapper;
+
+    public ErrorResponseFactory()
```
JanDoerrenhaus, line 18: I was a bit confused. At first I thought this change would force any user of mizool to now implement the `GlobalParametersSupplier` so that dependecy injection does not derp out. Then I realized that none of the classes affected by this PR partake in dependency injection. Now I began to wonder how one would even make use of this new feature, and I did not find that code anywhere.
Long story short, the only way to make use of this feature is by using neither the `AbstractErrorHandlingFilter`, nor the `RestExceptionMapper` from mizool, but roll your own mechanism. Using these two base features will in fact prevent you from making use of this new feature.

This smells. It should be possible - but not mandatory - to make use of this feature in the servlet and/or JAX-RS world, too.

JanDoerrenhaus (changes requested):

core/src/main/java/com/github/mizool/core/rest/errorhandling/ErrorMessageDto.java
```diff
@@ -7,23 +7,37 @@
...
         other.getErrors()
             .forEach(combined::putAll);
+        return combined;
+    }
 
-        return toBuilder().errors(combined.asMap())
-            .build();
+    private Map<String, Object> combineGlobalParameters(ErrorMessageDto other)
+    {
+        Map<String, Object> combined = Maps.newHashMap();
```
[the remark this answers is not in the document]
schroenser: The `ImmutableMap` will throw an exception on `.build()` when `other.getGlobalParameters()` contains keys that are also contained in `globalParameters`. My goal was to mimic the code, that combines the errors. However, that code will simply add another `ErrorDto` to the collection stored for a particular key, whereas my code would replace the value for the key, so they are not really comparable to each other.

The only remaining question would be: do we want an exception during `.build()`? This code is currently only used by majestix, which in turn doesn't use `globalParameters` anyways. So throwing an exception might reveal code inconsistencies. On the other hand, throwing exceptions during exception handling will hide the original exception.

But I think I'll at least wrap the return value in an unmodifiable decorator, though. Let me know if you think the immutable map, together with the exception, is better.

core/src/main/java/com/github/mizool/core/rest/errorhandling/ErrorResponseFactory.java
schroenser: Yeah, that's true. I'm also not really sure how to tackle this as all three known usages of the `ErrorResponseFactory`, the `AbstractErrorHandlingFilter`, the `RestExceptionMapper` and the `SystemDataResponseEntityExceptionHandler` rely on `new` rather than any form of injection.

As far as I can see, the `AbstractErrorHandlingFilter` and the `RestExceptionMapper` could have used injection. I don't know why they didn't. For the `SystemDataResponseEntityExceptionHandler`, should be possible as well, as long as I register the `ErrorResponseFactory` as a `@Bean` in some `@Configuration` class.

The question would be the "possible but not mandatory" part. If we want to solve this with injection, we need to have an injection point, which in turn means, we need to provide at least a noop-implementation that can be specialized inside the project. If I remember correctly, JavaEE allows this with the `@Specializes` annotation. I would have to investigate whether this also works for Spring.

Also, due to this topic, I'm beginning to wonder whether it was a good idea to use mizool for a Spring project in the first place or whether we should at least have forked this into "mizool-spring" or something.... 🤔

core/src/main/java/com/github/mizool/core/rest/errorhandling/ErrorMessageDto.java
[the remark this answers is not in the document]
JanDoerrenhaus: You could stream the `entrySet` of both maps, concatenate the strings, then collect the resulting combined stream into `ImmutableMap.toImmutableMap`, passing in a third parameter that uses the newer value if the key is already set.

core/src/main/java/com/github/mizool/core/rest/errorhandling/ErrorResponseFactory.java
JanDoerrenhaus: The goal should actually be to have a core that is completely agnostic, and does nothing by itself. It can be used in SE, in EE, in Spring, in Microprofile, whatever. Because it is just plain java.

And then you have technologies for the specific environments, that wrap the generic logic contained inside the core into the specifica needed in this environment.

schroenser: Good point. Seeing that among others, there are the following dependencies (among others) in `mizool-core`:
```xml
<dependency>
    <groupId>javax.enterprise</groupId>
    <artifactId>cdi-api</artifactId>
    <version>1.2</version>
    <scope>provided</scope>
</dependency>
<dependency>
    <groupId>javax.servlet</groupId>
    <artifactId>javax.servlet-api</artifactId>
    <version>3.1.0</version>
    <scope>provided</scope>
</dependency>
<dependency>
    <groupId>javax.ws.rs</groupId>
    <artifactId>javax.ws.rs-api</artifactId>
    <version>2.0.1</version>
</dependency>
```
I would say, we messed that goal up real good.

But that being said, I think the `ErrorResponseFactory` is not that far off as it _is_ plan Java. This leaves the question on how we want to change the `AbstractErrorHandlingFilter` and the `RestExceptionMapper`.
The `JacksonErrorHandlingFilter`, that extends the `AbstractErrorHandlingFilter`, already uses field injection, so I could try to change the `AbstractErrorHandlingFilter` to use field injection as well.
If I remember correctly, our resources in majestix also use field injection, so we might be lucky that the `RestExceptionMapper` can also use field injection.

We would have to pair that approach with a `NoopGlobalParametersSupplier` that can be specialized by the application.

schroenser: Since the `RestExceptionMapper` resides in `mizool-core`, I'll have to provide said supplier in the core as well.

schroenser: I now have made the changes locally, but I don't want to push them without first figuring out how to test them in a JavaEE environment. I'm currently not sure how to launch either majestix or any other of our EE environment. I'll have to figure that out tomorrow.

core/src/main/java/com/github/mizool/core/rest/errorhandling/GenericErrorMapper.java
```diff
@@ -65,13 +70,16 @@ private void logError(Throwable t, ErrorHandlingBehavior behavior)
...
         if (rootCause != t)
         {
-            behavior.getMessageLogLevel().log(log, "{} - {}", t.getMessage(), rootCause.getMessage());
+            behavior.getMessageLogLevel()
+                .log(log, "{} - {}", t.getMessage(), rootCause.getMessage());
         }
         else
         {
-            behavior.getMessageLogLevel().log(log, t.getMessage());
+            behavior.getMessageLogLevel()
```
JanDoerrenhaus, line 78: Where do all these weird short line breaks come from?

core/src/main/java/com/github/mizool/core/rest/errorhandling/ErrorMessageDto.java
```diff
@@ -2,26 +2,53 @@
...
+            .forEach(combined::putAll);
+        return Multimaps.unmodifiableSetMultimap(combined);
+    }
 
-        return toBuilder().errors(combined.asMap()).build();
+    private Map<String, Object> combineGlobalParameters(ErrorMessageDto other)
+    {
+        Map<String, Object> otherGlobalParameters = other.getGlobalParameters();
+
+        Set<String> combinedKeys = Sets.union(globalParameters.keySet(), otherGlobalParameters.keySet());
```
JanDoerrenhaus, line 43: I was actually thinking along the lines of streaming both key sets, then using `Stream.concat`. But this is acceptable as well.

core/src/main/java/com/github/mizool/core/rest/errorhandling/ErrorMessageDto.java
```diff
@@ -2,26 +2,53 @@
...
+
+        return combinedKeys.stream()
+            .collect(ImmutableMap.toImmutableMap(key -> key,
+                key -> resolveValueInOrder(key, otherGlobalParameters, globalParameters)));
```
JanDoerrenhaus, line 47: Ah, I see. It was important to have a key set free of duplicates. However, you could just have gone `ImmutableMap.toImmutableMap(key -> key, key -> globalParameters.getOrDefault(key, otherGlobalParameters.get(key)), BinaryOperators::first)`

As both approaches work, I leave it up to you whether to change this, or leave it as it is.

JanDoerrenhaus:
Check what's up with all these weird short line breaks. After that, I have no more comments, and the PR can be merged.

JanDoerrenhaus (changes requested):

core/src/main/java/com/github/mizool/core/rest/errorhandling/GenericErrorMapper.java
schroenser: In the spirit of "each operation should have it's own line", `Chained method calls: Wrap always` was enabled in the code style settings on [August 4, 2020](https://github.com/mizool/intellij-settings/commit/d4592b36261d415b3ab6eb3b00208f2f5110406c).

I think this was mainly intended for chained calls when working with builders or streams. Sadly the IntelliJ code formatter does not distinguish between different types of call chains.

schroenser:
fyi: @bannmann @gregorxiv

PR #202, opened 2021-12-03
Added interval helper class.

schroenser:


core/src/main/java/com/github/mizool/core/Interval.java
```diff
@@ -0,0 +1,44 @@
...
+ * }
+ * </pre>
+ */
+@RequiredArgsConstructor
+public class Interval
+{
+    private final Stopwatch stopwatch = Stopwatch.createStarted();
+    private final Duration duration;
+
+    public boolean isReached()
```
bannmann, line 33: It's highly unusual for a method called `is*()` (or any getter) to affect the state of the object. Even things like lazy instantiation, though they change the _internal_ state of an object, do not result in changed behavior as visible from the outside.

Also, the risk of not using this correctly is increased by the `isReached()` method not implying/ensuring via its parameters or name that you are supposed to perform an action if this returns `true`.

How about we let this method take a `Runnable` and rename it to `runIfDue`? The method can still return a boolean.

Also, we should either
- make this thread-safe (my preference given the usual use case) or
- document that it's not thread-safe - at least in Javadoc, but preferably also via a `@NotThreadSafe` annotation. (IIRC, we have one on the classpath as a transitive dependency of Guava.)

Finally, please add a Javadoc comment to the method, as well. It should mention that if the interval is _past_ due, a full interval is restarted (and not just for the remaining time), which has the effect that the points in time usually "shift". (Which is interesting because functionality like this could be intuitively expected to be like Cron, which will e.g. try to run the task again at the next regular scheduled time even though it was late for the last run.)

core/src/main/java/com/github/mizool/core/Interval.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.time.Duration;
+
+import lombok.RequiredArgsConstructor;
+
+import com.google.common.base.Stopwatch;
+
+/**
+ * Helper class to allow for something to happen after a certain amount of time. A common uses cases might be
+ * invalidation of in memory data.<br>
```
bannmann, line 11: fix typos: `use case` and `in-memory data`.

bannmann (changes requested):

core/src/main/java/com/github/mizool/core/Interval.java
schroenser: The idea with the runnable sounds nice on paper but it would affect the original use of this class, which was more like
```
if (!cache.contains(key) || cacheInvalidationInterval.isReached())
{
    refreshCache();
}
```

With the new design, that could only be refactored to
```
cacheInvalidationInterval.runIfDue(this::refreshCache);
if (!cache.contains(key))
{
    refreshCache();
}
```
which to me looks really weird.

Even with the `boolean` return value, this only burns down to
```
if (!cacheInvalidationInterval.runIfDue(this::refreshCache) && !cache.contains(key))
{
    refreshCache();
}
```
which I find even more confusing.

schroenser: Perhaps my impulse to label this class as "generic enough for mizool" was misguided, given that the project it's currently used in makes a lot of assumptions on how it's being used.

Feel free to decline this PR.

bannmann: Well, how about changing the application logic as follows?
```
// _clear_ the cache so that the subsequent condition is true
cacheInvalidationInterval.runIfDue(this::invalidateCache);

if (!cache.contains(key))
{
    refreshCache();
}
```

Only one `if` with no calls to methods which have side effects, only one `refreshCache()` call.

bannmann (approved):

PR #203, opened 2021-12-13
Add "unprocessable entity" exception types for common cases

bannmann:
Following up on #195, here are three exception types that can be used by store classes.

The goal is to avoid that each application defines similar exceptions (or worse, one set for each entity) for these common cases. 

IMO, their definition is sufficiently narrow to avoid them becoming the next "used for everything" exception that threatens the ability of applications to be properly internationalized.

## Notes

- I considered adding specialized constructors that take the field name, but ultimately decided against it: all three exception types cover cases that represent programming errors in a UI (or wrong usage when manually invoking REST APIs), so having placeholders and enabling error messages that say e.g. "Dear user, the field XXX should not be set for new entities" does not make much sense.
- The necessity to have these was discovered while working on the proof of concept code for our improved records helper.

### DIFF
--- a/core/src/main/java/com/github/mizool/core/exception/GeneratedFieldOverrideException.java
+++ b/core/src/main/java/com/github/mizool/core/exception/GeneratedFieldOverrideException.java
@@ -1,0 +1,28 @@
+package com.github.mizool.core.exception;
+
+/**
+ * Thrown when an attempt is made to set a value for a field which is automatically generated or maintained, such as the
+ * identifier or timestamps.
+ */
+public class GeneratedFieldOverrideException extends AbstractUnprocessableEntityException
+{
+    public GeneratedFieldOverrideException()
+    {
+        super();
+    }
+
+    public GeneratedFieldOverrideException(String message, Throwable cause)
+    {
+        super(message, cause);
+    }
+
+    public GeneratedFieldOverrideException(String message)
+    {
+        super(message);
+    }
+
+    public GeneratedFieldOverrideException(Throwable cause)
+    {
+        super(cause);
+    }
+}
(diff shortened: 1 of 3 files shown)


schroenser (changes requested):
I don't like the fact that you've decided against the field name.
While it's true, that we probably don't need the field name in the payload for user friendly messages, not having the field name makes finding a problem _way_ harder.

Experience has shown that a frontend developer can literally swamp the backend developers with questions like "why do I get `HTTP 422` when calling `PUT /x/x/z`? Oftentimes, they will even include the error response but most of the times, the backend dev _then_ has to ask for the original request body, just to compare it to the DTO, to figure out the reason.

I would suggest to add the `fieldName` as a `@NonNull` parameter and at least generating a helpful message in the constructor (which would mean that there are no constructors with `message` parameters).
I would even argue whether those exceptions need to be able to wrap a `cause` or whether they are most likely thrown due to a check and not re-thrown in a `catch` block.

The default behavior of the `AbstractUnprocessableEntityException` will contain the exception message, so at least a frontend developer has a chance to immediately spot the problem.

bannmann:
Good points! I added `fieldName` parameters to `GeneratedFieldOverrideException` and `ReadonlyFieldException`. Furthermore, I added `@Getter private final String fieldName;` to make the information available to potential exception behaviors.

I left `InvalidPrimaryKeyException` as it was because it needs configurable messages: "Primary key is null" and "Primary key mismatch". To me, it feels wasteful and overly specific to create two separate exceptions for that, especially because the primary key should be obvious enough that we don't need to point out its field name. Do you agree?

schroenser (approved):

PR #212, opened 2022-07-22
Sonar/20220722

schroenser:


core/src/main/java/com/github/mizool/core/validation/CheckDatabaseIdentifier.java
```diff
@@ -25,8 +29,9 @@ private boolean isValidValue(Object validationObject)
         if (validationObject instanceof String)
         {
             String validationString = (String) validationObject;
-            return !validationString.isEmpty() &&
-                validationString.matches("[a-zA-Z][a-zA-Z0-9_]*") &&
+            valid = !validationString.isEmpty() &&
```
bannmann, line 32: An empty string will not match the `[a-zA-Z]` part of `PATTERN`, so we can leave out the `isEmpty()` check.

bannmann: Please extend the unit test by adding the empty string to the `unacceptableValues` test data provider.

bannmann (changes requested):

bannmann (approved):

PR #213, opened 2022-08-11
Null safety in ErrorMessageDto.

JanDoerrenhaus:


schroenser (approved):
Note that Sonar complains about the third `if`, where the second half `otherGlobalParameters != null` is always true, due to the first `if`. In my opinion, this would be harder to read, if we performed that change.

core/src/main/java/com/github/mizool/core/rest/errorhandling/ErrorMessageDto.java
```diff
@@ -30,21 +30,45 @@ public ErrorMessageDto combineWith(ErrorMessageDto other)
...
+            result = otherGlobalParameters;
+        }
+        else
+        {
+            Set<String> combinedKeys = Sets.union(globalParameters.keySet(), otherGlobalParameters.keySet());
 
-        Set<String> combinedKeys = Sets.union(globalParameters.keySet(), otherGlobalParameters.keySet());
+            result = combinedKeys.stream()
+                .collect(ImmutableMap.toImmutableMap(key -> key,
+                    key -> resolveValueInOrder(key, otherGlobalParameters, globalParameters)));
```
bannmann, line 68: Please rename lambda arg `key` to `value`.

(Yes, this was already wrong before.)

bannmann:
> Note that Sonar complains about the third `if`, where the second half `otherGlobalParameters != null` is always true, due to the first `if`. In my opinion, this would be harder to read, if we performed that change.

I agree, let's not follow Sonar's advice here.

We *could* probably get rid of all the `if`s by always going through the `combinedKeys` approach. We would only need to do `if (parameters == null) parameters = Map.of();` for both maps at the start of the method. But I'm fine with leaving it like it is.

bannmann (changes requested):

core/src/main/java/com/github/mizool/core/rest/errorhandling/ErrorMessageDto.java
JanDoerrenhaus: I don't get it. We are streaming over combined keys, so does the lambda arg name `key` not make perfect sense here?

schroenser: Yes it does. :D Both the key function and the value function of the map collector get the `key` as parameter.

bannmann: Agreed, I was confused.

bannmann (approved):

PR #222, opened 2023-03-29
More Synchronizer options

bannmann:
### Motivation & background

Extend the Synchronizer API for usage scenarios where there is no specific "handover" by another thread that wakes up the synchronizer chains. One example is implementing scheduled polling that _can_, but _does not have to_ be quickened by explicit wake calls.

The code should be stable enough as it has been in production use for a while now.

### New features

1. Ability to specify an interval for sleeping after which the condition is checked again even when no wake call happens.
2. Allow sleeping as the sole action of a synchronizer chain.
3. Allow waking others as the sole action of a synchronizer chain.

In the diagram below, features 2 and 3 are highlighted (the updated diagram for Javadoc includes them but without highlighting):

synchronizer-api-change-illustration

### Other improvements

Additionally, this PR enables IntelliJ IDEA to complain when one forgets the final `invoke()` on a method chain of the fluent API:
``` java
synchronizer.sleepUntil(...)
    .run(...)
    .andWakeOthers(); // <- IDE warning: "Result of 'Synchronizer.andWakeOthers()' is ignored"
```

``` java
synchronizer.sleepUntil(...)
    .run(...)
    .andWakeOthers() // <- No warning now.
    .invoke();
```

Furthermore, javadocs are vastly improved.

Lastly, we now have isolated unit tests instead of only testing `Synchronizer` indirectly via `BufferedStreamAdapter`.

### Review hints

Reading the individual commits instead of the entire PR is recommended, starting with 7b5b92f894042a72c507c517e58a783136927286 "Add support for sleep timeout".

### DIFF
--- a/core/src/main/java/com/github/mizool/core/concurrent/SynchronizerApi.java
+++ b/core/src/main/java/com/github/mizool/core/concurrent/SynchronizerApi.java
@@ -6,32 +6,26 @@ import java.util.function.Predicate;
 import java.util.function.Supplier;
 
 /**
- * Provides fluent syntax for for {@link Synchronizer}.
+ * Provides fluent syntax for {@link Synchronizer}.
  */
 public interface SynchronizerApi
 {
-    /*
-     * Note on terminology: this API intentionally uses the verbs "sleep" and "wake". If it used "wait" and "notify"
-     * instead, its methods would all too easily be confused with methods in java.lang.Object which, if invoked on the
-     * objects returned by chained methods, could cause deadlocks.
-     */
-
     interface SleepRunGet extends RunGet
     {
         /**
          * Adds sleeping to the action chain. <br>
          * <br>
-         * When performing this action, the chain will ensure the given condition is met before performing the main
+         * When performing this action, the chain will ensure the given condition is met before performing the next
          * action. If the supplier returns {@code false}, this action chain sleeps until another chain wakes it up.
          * Then, the supplier is called again and if it still returns {@code false}, the chain resumes sleeping.
-         * Otherwise, the chain continues by performing the main action.
+         * Otherwise, the chain proceeds to perform the next action.
          *
          * @param state the supplier that returns {@code true} if the action chain should continue, {@code false}
          * otherwise.
          *
          * @throws NullPointerException if {@code state} is null
          */
-        default RunGet sleepUntil(BooleanSupplier state)
+        default RunGetInvoke sleepUntil(BooleanSupplier state)
         {
             return sleepUntil(state, null);
         }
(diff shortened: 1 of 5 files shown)


bannmann:
`testConditionallyNotWakingOthers()` failed and can't be fixed. It was built on the assumption that if there's no notify call, a waiting thread will never wake up. It turns out that this is wrong: the coding guideline mentioned in the code, [SEI CERT rule THI03-J](https://wiki.sei.cmu.edu/confluence/display/java/THI03-J.+Always+invoke+wait%28%29+and+await%28%29+methods+inside+a+loop), explains that something called "spurious wakeup" can in fact happen to any waiting thread. I even observed one on my machine: the line in `Synchronizer` calling `notifyAll()` was never reached, but still the thread woke up. 

THI03-J says that developers just have to wrap the waiting in a loop that checks a condition. And that's exactly what `Synchronizer` does; it even encapsulates that so the developer can't do anything wrong! Unless, of course, the developer in question happens to be writing a unit test for `Synchronizer` itself, creating an artifical scenario where it's important whether the _condition_ is evaluated or not. 🤦‍♂️

Real code would not have this problem as it only cares about a `Synchronizer` chain's main action (`run()` or `get()`). And that action _is_ properly guarded against spurious wakeups _by the condition_!

core/src/main/java/com/github/mizool/core/concurrent/SynchronizerApi.java
```diff
@@ -20,16 +21,37 @@ interface SleepRunGet extends RunGet
...
+         * chain resumes sleeping. Otherwise, the chain continues by performing the main action.
+         *
+         * @param state the supplier that returns {@code true} if the action chain should continue, {@code false}
+         * otherwise.
+         * @param timeout the maximum amount of time to sleep before re-checking the condition, or {@code null} to
+         * sleep indefinitely.
+         *
+         * @throws NullPointerException if {@code state} is null
+         */
+        RunGet sleepUntil(BooleanSupplier state, Duration timeout);
```
schroenser: Would it make sense to exchange the boolean supplier for a function that receives the information about whether the timeout was reached or another thread was waking it up? Something like `Function<WakeReason, Boolean>` perhaps?

Since the second parameter is called `timeout`, I would imagine that implementations of `state` might want to act differently, depending on whether the process was resumed "normally" or went into a timeout.

schroenser (approved):

core/src/main/java/com/github/mizool/core/concurrent/SynchronizerApi.java
schroenser: @bannmann I've approved and the pull request as I don't see this as something we _have_ to do but more as a conversation starter.

If you want' we can talk about this. But "yagni" is also fine. ;)

bannmann: For the record: we discussed this in person and the tl;dr of our conclusion is YAGNI.

PR #224, opened 2023-05-05
Improve Synchronizer API docs

bannmann:


bannmann:
Skipping formal review as these are only API doc improvements.

PR #232, opened 2024-04-25
minor dependency updates

voons:


pom.xml
```diff
@@ -229,27 +229,27 @@
                 <scope>import</scope>
             </dependency>
             <dependency>
-                <!-- force dependency convergence. 1.1.3 is intended for aws-java-sdk-dynamodb 1.12.569 -->
```
bannmann, line 232: These "force dependency convergence" should mention both relevant versions - you updated the version that this one is taken from, but did not change `1.1.3` to `1.15`. It seems this was not always done previously, either.

We should definitely have a discussion about dependency convergence in general; maybe striving for convergence is not worth the hassle after all.

Anyway, until we potentially abolish or change the practice, please make all these comments consistent with the respective versions.

bannmann (changes requested):

bannmann (approved):

bannmann:
FYI @JanDoerrenhaus @derWhity

PR #237, opened 2024-06-19
Feature/updated minor dependencies

voons:


pom.xml
```diff
@@ -294,10 +295,10 @@
                 <version>1.1.1</version>
             </dependency>
             <dependency>
-                <!-- possible dependency convergence problem with transitive dependencies of mockito-core (1.14.8) and assertj-core (1.14.13)-->
+                <!-- possible dependency convergence problem with transitive dependencies of mockito-core (1.14.16) and assertj-core (1.14.13)-->
```
bannmann, line 298: Such comments won't age well unless we also list the respective direct versions involved, please change this as follows:

```suggestion
                <!-- possible dependency convergence problem with transitive dependencies of mockito-core 5.12.0 (1.14.16) and assertj-core 3.26.0 (1.14.13)-->
```

bannmann (changes requested):

bannmann (approved):

bannmann:
FYI @JanDoerrenhaus @derWhity

PR #260, opened 2025-09-12
Add FluentSynchronizer, deprecate Synchronizer for removal

bannmann:
`Synchronizer` was originally built to power `BufferedStreamAdapter`, but has since been used in several downstream projects. Recently, it became clear in one of these that Synchronizer's API is not flexible enough as it was built around the idea of each chain having only one main action - `get()` or `run()`.

This PR introduces a replacement class `FluentSynchronizer`, which lifts that limitation and allows chains like the following (note that there are both `run()` and `get()` calls):

``` java
boolean receivedFooWhileWaiting = synchronizer.define()
    .run(() -> {
        // ...
        receivedFoo = false;
    })
    .sleepUntil(() -> receivedFoo, Duration.ofSeconds(10))
    .get(() -> receivedFoo)
    .invoke();

// if receivedFooWhileWaiting is false, there was a timeout
```
_[Edit: the simplified example above won't work properly as-is. The `sleepUntil()` condition in the original code also checks for a deadline.]_

The new API is based on the old one, but makes significant changes:
- `run()` and `get()` can be called multiple times
- Method name prefixes `and` & `then` were dropped
- Action chains now always start with a `define()` call
    - In contrast, `Synchronizer` has multiple entry methods which makes it more cumbersome to add/change action methods)
- The term _chain result_ was introduced to avoid having to say "if `get()` was called" or "the result of the last `get()` action"
- A method to map the chain result (similar to `Optional.map()`)
- A method to explicitly discard the chain result
- Javadoc was improved, parts were rewritten
- The new implementation is shorter and hopefully more straightforward

Note that `FluentSynchronizerTest` is based on (and functionally identical to) `SynchronizerTest`, so this rewrite likely did not introduce bugs to existing functionality.

### DIFF
--- a/core/src/main/java/com/github/mizool/core/concurrent/BufferedStreamAdapter.java
+++ b/core/src/main/java/com/github/mizool/core/concurrent/BufferedStreamAdapter.java
@@ -197,6 +197,7 @@ public final class BufferedStreamAdapter<F, V>
     private final class BlockingSpliterator implements Spliterator<V>
     {
         @Override
+        @SuppressWarnings("java:S2589") // Sonar thinks valueHolder local variable is non-nullable
         public boolean tryAdvance(Consumer<? super V> action)
         {
             ValueHolder<V> valueHolder = synchronizer.define()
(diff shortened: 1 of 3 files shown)


lesmou (approved):
My sonar lint plugin had some wild recommendations concerning "useless assignments" in FluentSynchronizers get and map functions but I guess you saw them.
Moreover nothing to complain about.
Overall I am lacking detailed knowledge about the field of application to provide a more in depth review.

